PR #11, opened 2020-09-09
Added Metrics and losses(Proposal)

ravinkohli:


autoPyTorch/pipeline/components/training/losses.py
```diff
@@ -0,0 +1,59 @@
...
+    BCEWithLogitsLoss,
+    MSELoss,
+    L1Loss
+)
+from torch.nn.modules.loss import _Loss as Loss
+
+
+losses = dict(classification=dict(
+    CrossEntropyLoss=dict(
+        module=CrossEntropyLoss, supported_output_type='multi-class'),
```
franchuterivera, line 14: Ideally, we should have a single mapping like https://github.com/automl/auto-sklearn/blob/7a3f3a517821da7f00d92991960dd474a22c223e/autosklearn/constants.py#L16

which we can use here. I think @bastiscode  was creating something like this... maybe you can use what he is using?

autoPyTorch/pipeline/components/training/losses.py
```diff
@@ -0,0 +1,59 @@
...
+    BCEWithLogitsLoss=dict(
+        module=BCEWithLogitsLoss, supported_output_type='binary-class')),
+    regression=dict(
+        MSELoss=dict(
+            module=MSELoss, supported_output_type='continuous'),
+        L1Loss=dict(
+            module=L1Loss, supported_output_type='continuous')))
+
+default_losses = dict(classification=CrossEntropyLoss, regression=MSELoss)
```
franchuterivera, line 23: I think this will change, so that binary classification is a task. That means, we are gonna need a default loss for this case.

autoPyTorch/pipeline/components/training/metrics/base_metric.py
```diff
@@ -0,0 +1,25 @@
+from typing import Any, Dict, Optional
+
+from pytorch_lightning.metrics.metric import Metric
+
+from sklearn.base import BaseEstimator
+
+import torch.tensor
+
+
+class autoPyTorchMetric(BaseEstimator):
```
franchuterivera, line 10: Do you think you can add a comment saying why you use base estimator? I think I recall you wanted to use the get params set params framework, but for future readers, It will be more understandable if there is a comment explaining that

autoPyTorch/pipeline/components/training/metrics/base_metric.py
```diff
@@ -0,0 +1,25 @@
...
+    def __init__(self):
+        self.metric: Optional[Metric] = None
```
franchuterivera, line 12: I think we need a mechanism to ensure that the optimizer (BOHB, SMAC, etc) can tell if this metric is better the smaller it gets or better if it goes higher.

Could you add an attribute like that?

test/test_pipeline/test_metrics.py
```diff
@@ -0,0 +1,47 @@
+import unittest
+
+import torch
+
+from autoPyTorch.pipeline.components.training.metrics.base_metric import autoPyTorchMetric
+from autoPyTorch.pipeline.components.training.metrics.utils import get_metrics
+from autoPyTorch.pipeline.components.training.metrics.utils import get_supported_metrics
+
+
+class MetricsTest(unittest.TestCase):
```
franchuterivera, line 10: When you add an attribute to tell if a loss is better the higher it gets, or the lower it gets, we should add a test for this.

autoPyTorch/pipeline/components/training/metrics/SSIM.py
```diff
@@ -0,0 +1,46 @@
...
+    def __call__(self,
+                 predictions: torch.tensor,
+                 targets: torch.tensor
+                 ) -> torch.tensor:
+        return self.metric(predictions, targets)
+
+    @staticmethod
+    def get_properties(dataset_properties: Optional[Dict[str, Any]] = None) -> Dict[str, str]:
+        return {
+            'shortname': 'Metric',
```
LMZimmer, line 43: This should be something else, I think ;)

autoPyTorch/pipeline/components/training/losses.py
LMZimmer: I agree, he is on vacation this week though

autoPyTorch/pipeline/components/training/losses.py
LMZimmer: ... which should also be CE then

autoPyTorch/pipeline/components/training/metrics/AUROC.py
```diff
@@ -0,0 +1,37 @@
...
+    def __call__(self,
+                 predictions: torch.tensor,
+                 targets: torch.tensor
+                 ) -> torch.tensor:
+        return self.metric(predictions, targets)
+
+    @staticmethod
+    def get_properties(dataset_properties: Optional[Dict[str, Any]] = None) -> Dict[str, str]:
+        return {
+            'shortname': 'AUROC',
```
LMZimmer, line 34: I think this could just be AUC

autoPyTorch/pipeline/components/training/metrics/Accuracy.py
```diff
@@ -0,0 +1,37 @@
...
+                 targets: torch.tensor
+                 ) -> torch.tensor:
+        return self.metric(predictions, targets)
+
+    @staticmethod
+    def get_properties(dataset_properties: Optional[Dict[str, Any]] = None) -> Dict[str, str]:
+        return {
+            'shortname': 'Acc',
+            'name': 'Accuracy',
+            'task_type': 'classification'
```
LMZimmer, line 36: Does this work if some metrics are usable for mutliple task types, e.g. binary_classification, multi_classification?

autoPyTorch/pipeline/components/training/metrics/RMSLE.py
```diff
@@ -0,0 +1,31 @@
...
+from pytorch_lightning.metrics import regression
+from pytorch_lightning.metrics.metric import Metric
+
+import torch.tensor
+
+from autoPyTorch.pipeline.components.training.metrics.base_metric import autoPyTorchMetric
+
+
+class RMSLE(autoPyTorchMetric):
+    def __init__(self,
```
LMZimmer, line 12: Could you add a description of how the log transform is applied?

autoPyTorch/pipeline/components/training/metrics/base_metric.py
ravinkohli: in the end, I just kept it to keep it consistent with the rest of the code. I dont know if we'll use get_params or set_params. I can totally change it to inherit from object.

autoPyTorch/pipeline/components/training/losses.py
ravinkohli: So, by single mapping you mean the same for loss and metrics? I didnt quite understand what you mean, could you explain further on where this mapping is to be used?

autoPyTorch/pipeline/components/training/losses.py
ravinkohli: I have made it in a way that all types of classification have crossentropy as the default and all types of regression have MSE,  so what would happen is the task type comming from dataset properties would be split, [like here](https://github.com/ravinkohli/Auto-PyTorch_refactor/blob/e1182883ddc8806bd98b2e93c12aebbe3df80f7a/autoPyTorch/pipeline/components/training/losses.py#L27).

autoPyTorch/pipeline/components/training/metrics/Accuracy.py
ravinkohli: yes, this is handled similar to how loss is handled (explained in a comment above).

autoPyTorch/pipeline/components/training/losses.py
franchuterivera: Of course. We all have to speak the same language. 

If I ask in my part of the code for 'binary_classification' and you in your code, you ask for 'BINARYCLASSIFICATION' every time you/me have the need to query this information, it becomes tedious. 

Can you please create a file named constants.py like the one above, where you use an enumerate so we standardized the question:
```
If BINARY_CLASSIFICATION:
    # Do Something
```
Where BINARY_CLASSIFICATION will be a enum object in this file that you create

autoPyTorch/pipeline/components/training/metrics/base_metric.py
franchuterivera: Yeah, any place you see the possibility to save memory, please go with it. Eventually, it will be great if we compare our ram usage to see if we have room for improvement.

ravinkohli: okay, will do

autoPyTorch/constants.py
```diff
@@ -0,0 +1,40 @@
...
+    {TABULAR_CLASSIFICATION: 'tabular_classification',
+     IMAGE_CLASSIFICATION: 'image_classification',
+     TABULAR_REGRESSION: 'tabular_regression',
+     IMAGE_REGRESSION: 'image_regression'}
+
+STRING_TO_TASK_TYPES = \
+    {'tabular_classification': TABULAR_CLASSIFICATION,
+     'image_classification': IMAGE_CLASSIFICATION,
+     'tabular_regression': TABULAR_REGRESSION,
+     'image_regression': IMAGE_REGRESSION}
```
franchuterivera, line 21: I think we can later enhance the one for images, like adding detection and so on. I am happy with the tabular ones, which will let us test the pipeline.

franchuterivera:
All of my comments have been addressed. @LMZimmer  If you are ok with these changes, can we squash and merge this?

PR #14, opened 2020-09-22
Image Pipeline

ravinkohli:
1. Split preprocessing pipeline into tabular and image preprocessing
2. added preprocessing node
3. added Image classification pipeline with example on fashion-mnist


autoPyTorch/pipeline/components/preprocessing/image_preprocessing/padding/Pad.py
```diff
@@ -0,0 +1,59 @@
...
+)
+
+import numpy as np
+
+import torch.tensor
+
+from autoPyTorch.pipeline.components.preprocessing.image_preprocessing.padding.base_pad import BasePad
+
+
+class Pad(BasePad):
```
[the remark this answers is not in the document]
franchuterivera: I see, yeah I think it makes sense to have it as an augmentation strategy, not so much as pre-processing. 

Ravin, what do you think? I am talking about 
https://github.com/LMZimmer/Auto-PyTorch_refactor/blob/08e319cc0bb1c41af376e765c6d5155924cae94e/autoPyTorch/pipeline/image_classification.py#L189

That is, instead of a fixed step in the pipeline, does it makes sense to have it as a augmentation choice?

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/utils.py
```diff
@@ -0,0 +1,17 @@
+from typing import Any, Dict, List
+
+from sklearn.base import BaseEstimator
+
+
+def get_tabular_preprocessers(X: Dict[str, Any]) -> Dict[str, List[BaseEstimator]]:
```
franchuterivera, line 6: Do you think you can add a description saying:

We expect that the fit dictionary would have previously fitted the numerical/categorical nodes that will build  the column transformer of the tabular preprocessing component. This function parses the fict dictionary and extract such components.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/utils.py
```diff
@@ -0,0 +1,17 @@
...
+    preprocessor = dict(numerical=list(), categorical=list())  # type: Dict[str, List[BaseEstimator]]
+
+    for key, value in X.items():
+        if isinstance(value, dict):
+            if 'numerical' or 'categorical' in value.keys():
+                if isinstance(value['numerical'], BaseEstimator):
```
franchuterivera, line 12: Instead of asking for <BaseEstimator>, doesn't it makes sense to ask for <autoPyTorchPreprocessingComponent> component, just to be safe?

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/utils.py
```diff
@@ -0,0 +1,17 @@
...
+                    preprocessor['numerical'].append(value['numerical'])
+                if isinstance(value['numerical'], BaseEstimator):
```
franchuterivera, line 14: same here

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/TabularColumnTransformer.py
```diff
@@ -0,0 +1,87 @@
...
+        Args:
+            X (Dict[str, Any]): fit dictionary
+
+        Returns:
+            "TabularColumnTransformer": an instance of self
+        """
+        self.check_requirements(X, y)
+
+        preprocessors = get_tabular_preprocessers(X)
+        numerical_pipeline = make_pipeline(*preprocessors['numerical'])
```
franchuterivera, line 48: If the preprocessors['numerical'] is empty, because the data is only categorical, will this still work?

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/encoding/OneHotEncoder.py
```diff
@@ -2,27 +2,24 @@
...
 
-from autoPyTorch.pipeline.components.preprocessing.encoding.base_encoder import BaseEncoder
+from autoPyTorch.pipeline.components.preprocessing.tabular_preprocessing.encoding.base_encoder import BaseEncoder
 
 
 class OneHotEncoder(BaseEncoder):
     """
     Encode categorical features as a one-hot numerical array
     """
     def __init__(self, random_state: Optional[Union[np.random.RandomState, int]] = None):
```
franchuterivera, line 14: I see this repeated in the NoEncoder and here. Do you think it makes sense to move it to the base class?

autoPyTorch/pipeline/components/setup/preprocessor/Preprocessing.py
```diff
@@ -0,0 +1,69 @@
...
+
+from ConfigSpace.configuration_space import ConfigurationSpace
+
+import numpy as np
+
+from autoPyTorch.pipeline.components.setup.base_setup import autoPyTorchSetupComponent
+from autoPyTorch.pipeline.components.setup.preprocessor.utils import get_preprocess_transforms, preprocess
+
+
+class Preprocessing(autoPyTorchSetupComponent):
```
franchuterivera, line 11: Do you think you can change the name to something that indicates that preprocessing may happen here?
Ideas:

- EarlyPreprocessing
-PrecedentPreprocessing

Yeah or something nice, but that indicates that this class will do an early preprocessing for efficiency :)

requirements.txt
```diff
@@ -1,4 +1,8 @@
 torch
 ConfigSpace
-scikit-learn>=0.22.0,<0.23
+scikit-learn
```
franchuterivera, line 3: Do you think we can stick with scikit learn earlier than 0.23? There are changes in 0.23 that we haven't fully tested here not in auto-sklearn

requirements.txt
```diff
@@ -1,4 +1,8 @@
...
 pandas
+torchvision
+pytorch-lightning
```
franchuterivera, line 6: Sorry to ask, because I thought we discussed it before and maybe I misunderstood. 

I understood @LMZimmer recommended not having this dependency? 

But again, maybe I missunderstood.

autoPyTorch/pipeline/components/setup/preprocessor/utils.py
```diff
@@ -0,0 +1,25 @@
...
+        if isinstance(value, autoPyTorchPreprocessingComponent):
+            transforms.append(value)
+
+    return torchvision.transforms.Compose(transforms)
+
+
+def preprocess(dataset: np.ndarray, transforms: torchvision.transforms.Compose) -> np.ndarray:
+    # uncomment next 2 lines if dataset is not np array
+    # dataset = copy.copy(dataset)
+    # dataset.data = transforms(dataset.data)
```
franchuterivera, line 22: This looks like tensor, so it makes me wonder... instead of having this hear, does it make sense to create an issue to support gpu preprocessing later?

autoPyTorch/pipeline/components/setup/preprocessor/Preprocessing.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def __init__(self, random_state: Optional[np.random.RandomState] = None):
+        self.random_state = random_state
+
+    def fit(self, X: Dict[str, Any], y: Any = None) -> "Preprocessing":
```
franchuterivera, line 16: Shouldn't the preprocessing happen here? I think that fit_transform flow goes like this:

1. fit is called in every component
2. after fit (and here it means we even fitted the network) this preprocessing transform will be called

Sof for the preprocessing to be called I think we need to call  X['train'] = preprocess(dataset=X['train'], transforms=transforms) here

We also need to call preprocessing in the validation data. but I will add a commend on that at the end

autoPyTorch/pipeline/image_classification.py
```diff
@@ -0,0 +1,206 @@
...
+                by the pipeline.
+        """
+        steps = []  # type: List[Tuple[str, autoPyTorchChoice]]
+
+        default_dataset_properties = {'target_type': 'image_classification'}
+        if dataset_properties is not None:
+            default_dataset_properties.update(dataset_properties)
+
+        steps.extend([
+            ("pad", PadChoice(default_dataset_properties)),
```
franchuterivera, line 189: I think probably padding makes sense as augmentation?

autoPyTorch/pipeline/tabular_classification.py
```diff
@@ -187,6 +193,8 @@ def _get_pipeline_steps(self, dataset_properties: Optional[Dict[str, Any]],
             ("imputer", SimpleImputer()),
             ("encoder", EncoderChoice(default_dataset_properties)),
             ("scaler", ScalerChoice(default_dataset_properties)),
+            ("tabular_transformer", TabularColumnTransformer()),
+            ("preprocessing", Preprocessing()),
```
franchuterivera, line 197: here we also have to update the name

test/test_pipeline/components/test_tabular_column_transformer.py
```diff
@@ -0,0 +1,72 @@
...
+        # check if transformer was added to fit dictionary
+        self.assertIn('tabular_transformer', X.keys())
+        # check if transformer is of expected type
+        self.assertIsInstance(column_transformer, ColumnTransformer)
+
+        data = column_transformer.fit_transform(X['train'])
+        self.assertIsInstance(data, np.ndarray)
+
+
+if __name__ == '__main__':
```
franchuterivera, line 71: I think you  should probably add a test for only categorical only numerical data

franchuterivera (changes requested):
I think the test are failing, so we have to look into the flake errors 

Also, some small enhancements where requested. What I think we need to add tho, is the support for validation and train data.

That is:

- I think we will have 3 types of data. X_train, X_val (always there) and and optional X_test. All should go through preprocessing. Can you please add this change also?

For example:

https://github.com/LMZimmer/Auto-PyTorch_refactor/blob/08e319cc0bb1c41af376e765c6d5155924cae94e/autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/TabularColumnTransformer.py#L54

from:
```
self.column_transformer.fit(X['train'])
```

to:
```
self.column_transformer.fit(X['X_train'])
self.column_transformer.fit(X['X_val'])
if 'X_test' in X:
    self.column_transformer.fit(X['X_test'])
```

requirements.txt
ravinkohli: dependency on pytorch lightning? As far as I remember we decided to have it as a dependency.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/utils.py
ravinkohli: now the tabular preprocessors are adding the sklearn preprocessors to the fit dictionary. Therefore the only base class they share is base estimators

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/TabularColumnTransformer.py
ravinkohli: I thought I made it work but apparently it doesn't so I will fix it and add a test to ensure only numerical and only categorical cases work

autoPyTorch/pipeline/components/setup/preprocessor/Preprocessing.py
ravinkohli: but then as all the components are added to the fit_dictionary by calling transform on the node, so I think there would not be any preprocessors in the fit_dictionary if this preprocess is done in the fit function.

autoPyTorch/pipeline/image_classification.py
ravinkohli: okay then I'll move it to augmentation when I make a PR for that.

autoPyTorch/pipeline/components/preprocessing/image_preprocessing/padding/Pad.py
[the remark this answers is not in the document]
ravinkohli: In case of augmentation, imgaug already has other ways of doing something like the RandomCrop plus pad(they have a crop plus keep size method). I put pad in preprocessing as if it is to be done on the images independently then it should be done as a preprocessing module(as then it will save time). I think we dont need pad as a node if wee just want the random crop plus padding functionality.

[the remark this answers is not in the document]
franchuterivera: yeah random crop plus padding sounds great, in the augmentation context

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/utils.py
franchuterivera: Oh, I see.. then It makes sense. Just please add a comment in code explaining this.

requirements.txt
franchuterivera: Oh, I see. Cool, then I will mark this as resolved

autoPyTorch/pipeline/components/setup/preprocessor/Preprocessing.py
franchuterivera: Yeah, please ignore this. I missread:
https://github.com/scikit-learn/scikit-learn/blob/0fb307bf39bbdacd6ed713c00724f8f871d60370/sklearn/pipeline.py#L742

But you are entirely right. I will mark this as resolved

franchuterivera (approved):

franchuterivera:
some mypy test was still failing. Also the example for tabular classification was not working, but all minor changes. So I added the changes myself to expedite the process. All comments have been addressed so I am merging this now!

PR #18, opened 2020-10-25
Image augmentation

ravinkohli:
Added:

1. image augmentations
2. ImageAugmenter that aggregates the various augmentations automatically and create a sequential
3. ImageDataLoader that can add image augmenter to transforms


autoPyTorch/pipeline/components/setup/early_preprocessor/EarlyPreprocessing.py
```diff
@@ -67,8 +67,8 @@ def get_hyperparameter_search_space(
     @staticmethod
     def get_properties(dataset_properties: Optional[Dict[str, Any]] = None) -> Dict[str, str]:
         return {
-            'shortname': 'Preprocessing',
-            'name': 'Preprocessing Node',
+            'shortname': 'EarlyPreprocessing',
```
franchuterivera, line 70: haha my PR and your PR will class on this change. Depending on which get merged first, this will have to be updated

autoPyTorch/pipeline/components/setup/augmentation/image/base_image_augmenter.py
```diff
@@ -0,0 +1,53 @@
...
+        has been called on the object.
+        Returns:
+            BaseEstimator: Fitted augmentor
+        """
+        if self.augmenter is None:
+            raise AttributeError("Can't return augmenter for {}, as it has not been"
+                                 " fitted yet".format(self.__class__.__name__))
+        return self.augmenter
+    #
+    # def check_requirements(self, X: Dict[str, Any], y: Any = None) -> None:
```
franchuterivera, line 29: Why is this commented out :D?

autoPyTorch/pipeline/components/setup/augmentation/image/ImageAugmenter.py
```diff
@@ -0,0 +1,135 @@
...
+        available_augmenters = get_components()  # type: Dict[str, BaseImageAugmenter]
+
+        if dataset_properties is None:
+            dataset_properties = dict()
+
+        # add child hyperparameters
+        for name in available_augmenters.keys():
+            preprocessor_configuration_space = available_augmenters[name].\
+                get_hyperparameter_search_space(dataset_properties)
+            cs.add_configuration_space(name, preprocessor_configuration_space)
```
franchuterivera, line 118: I fail to understand something -- let us say  we don't want cropping for this pipeline because it hurts performance. Is there the possibility to not have cropping? Like normally I will expect a 'no cropping' or something similar. Maybe I missed something :)

autoPyTorch/pipeline/components/setup/augmentation/image/ImageAugmenter.py
```diff
@@ -0,0 +1,135 @@
...
+
+class ImageAugmenter(BaseImageAugmenter):
+
+    def __init__(self, random_state: Optional[Union[int, np.random.RandomState]] = None):
+        super().__init__()
+        self.available_augmenters = get_components()  # type: Dict[str, BaseImageAugmenter]
+        self.random_state = random_state
+
+    def fit(self, X: Dict[str, Any], y: Any = None) -> BaseImageAugmenter:
+        self.augmenter = iaa.Sequential([augmenter.fit(X).get_image_augmenter() for _, augmenter in
```
franchuterivera, line 56: Did autopytorch support object detection? else, we will have to consider something like https://imgaug.readthedocs.io/en/latest/source/examples_bounding_boxes.html

test/test_pipeline/components/test_setup_image_augmenter.py
```diff
@@ -0,0 +1,46 @@
+import unittest
+
+from imgaug.augmenters.meta import Augmenter, Sequential
+
+import numpy as np
+
+from autoPyTorch.pipeline.components.setup.augmentation.image.ImageAugmenter import ImageAugmenter
+
+
+class TestImageAugmenter(unittest.TestCase):
```
franchuterivera, line 10: Can you add a test that actually takes an image and apply a transformation, to make sure it properly happen? It can be that it takes a random transformation on a manually created 1,H,W,C numpy array and when applied we make sure some change happened. This will also prevent bugs on theimplementation

franchuterivera (changes requested):
I left some comments that will likely request a change. Please take a look and see if they make sense

autoPyTorch/pipeline/components/setup/augmentation/image/base_image_augmenter.py
ravinkohli: I'm sorry I had copied first in case I need requirements. As augmenter has no requirements, it is not needed but I forgot to delete

autoPyTorch/pipeline/components/setup/augmentation/image/ImageAugmenter.py
ravinkohli: So, the idea is for augmenters to have a probability of being applied as a hyperparameter. According to Lucas' comments, currently there is one just for RandomCutout. But it can be added for others as well.

autoPyTorch/pipeline/components/training/metrics/utils.py
```diff
@@ -32,7 +32,7 @@ def get_components() -> Dict[str, autoPyTorchMetric]:
             as choices
     """
     components = OrderedDict()
-    components.update(_encoders)
+    components.update(_metrics)
```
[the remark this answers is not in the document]
ravinkohli: Okay

autoPyTorch/pipeline/components/setup/augmentation/image/ImageAugmenter.py
ravinkohli: As far as I am aware, autopytorch didnt suport object detection. I can change it to support object detection if we decide to have that task after refactoring

autoPyTorch/pipeline/components/training/data_loader/image_data_loader.py
```diff
@@ -0,0 +1,61 @@
...
+
+        if train:
+            transformations.append(X['image_augmenter'])
+        # In the case of image data, the options currently available
+        # for preprocessors are:
+        #   + normalise
+        # These can apply for both train/val/test, so no
+        # distinction is performed
+
+        if not X['is_small_preprocess']:
```
LMZimmer, line 38: Could you add documentation what this means?

LMZimmer (changes requested):
Looking good, could you also add a resize as an augmentation?

autoPyTorch/pipeline/components/training/data_loader/image_data_loader.py
ravinkohli: yeah sure

autoPyTorch/pipeline/components/training/metrics/utils.py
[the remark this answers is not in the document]
franchuterivera: Yes!

test/test_pipeline/components/test_setup_image_augmenter.py
```diff
@@ -0,0 +1,49 @@
...
+    def test_every_augmenter(self):
+        image_augmenter = ImageAugmenter()
+        configuration = image_augmenter.get_hyperparameter_search_space().sample_configuration()
+        image_augmenter = image_augmenter.set_hyperparameters(configuration=configuration)
+        X = dict(X_train=np.random.randint(0, 255, (8, 3, 16, 16), dtype=np.uint8), image_height=16, image_width=16)
+        for name, augmenter in image_augmenter.available_augmenters.items():
+            if not augmenter.use_augmenter:
```
LMZimmer, line 17: To test every augmenter, there should be a configuration that has `use_augmenter==True` for all augmenters right?

LMZimmer (changes requested):
Just a few comments, once these are addressed I will merge it 👍

autoPyTorch/pipeline/components/setup/augmentation/image/Resize.py
```diff
@@ -0,0 +1,69 @@
...
+
+    @staticmethod
+    def get_hyperparameter_search_space(
+        dataset_properties: Optional[Dict[str, str]] = None
+    ) -> ConfigurationSpace:
+
+        cs = ConfigurationSpace()
+        interpolation = CategoricalHyperparameter('interpolation', choices=['nearest', 'linear', 'area', 'cubic'],
+                                                  default_value='linear')
+        use_augmenter = CategoricalHyperparameter('use_augmenter', choices=[True, False])
```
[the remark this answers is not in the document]
ravinkohli: Actually I have made use_augmenter as an attribute of the base class. So, all of the augmenters have the same name

autoPyTorch/pipeline/components/setup/augmentation/image/ZeroPadAndCrop.py
```diff
@@ -0,0 +1,71 @@
...
+from autoPyTorch.pipeline.components.setup.augmentation.image.base_image_augmenter import BaseImageAugmenter
+
+
+class ZeroPadAndCrop(BaseImageAugmenter):
+    def __init__(self, x_position: float = 0, y_position: float = 0,
+                 percent: float = 0.1, random_state: Optional[Union[int, np.random.RandomState]] = None):
+        super().__init__()
+        self.random_state = random_state
+        self.percent = percent
+        self.position = (x_position, y_position)
```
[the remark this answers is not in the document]
ravinkohli: okay, so there is sampling from uniform or normal. I have set it to uniform but let me know if normal makes more sense.

autoPyTorch/pipeline/components/setup/augmentation/image/RandomAffine.py
```diff
@@ -0,0 +1,71 @@
...
+
+        cs = ConfigurationSpace()
+        scale_min = UniformFloatHyperparameter('scale_min', lower=0, upper=0.99, default_value=0)
+        scale_offset = UniformFloatHyperparameter('scale_offset', lower=0, upper=0.99, default_value=0.2)
+        translate_percent_min = UniformFloatHyperparameter('translate_percent_min', lower=0, upper=0.99,
+                                                           default_value=0)
+        translate_percent_offset = UniformFloatHyperparameter('translate_percent_offset', lower=0, upper=0.99,
+                                                              default_value=0.3)
+        shear = UniformIntegerHyperparameter('shear', lower=0, upper=45, default_value=30)
+        rotate = UniformIntegerHyperparameter('rotate', lower=0, upper=360, default_value=45)
```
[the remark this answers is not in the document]
ravinkohli: So I had made a mistake in interpreting what scale means. So, if we have scale as 1, that would mean scaled to 100% of original size(i.e, no change) else with 0.5 then 50% of the image. So I think scale offset should be 0 to 0.4(therefore scale_min) and it should go from scale_min to 1 and that should mean we scale it to anything between 60%(1-0.4) and 100%(1)

test/test_pipeline/components/test_setup_image_augmenter.py
```diff
@@ -10,12 +10,10 @@
 class TestImageAugmenter(unittest.TestCase):
     def test_every_augmenter(self):
         image_augmenter = ImageAugmenter()
-        configuration = image_augmenter.get_hyperparameter_search_space().sample_configuration()
+        configuration = image_augmenter.get_hyperparameter_search_space().get_default_configuration()
```
LMZimmer, line 13: Does the default have every augmentation? I guess then this is fine, but perhaps add a comment here that this is implicitly assumed for this test

autoPyTorch/pipeline/components/setup/augmentation/image/RandomAffine.py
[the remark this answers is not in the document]
LMZimmer: I would expect that if you specify the scaling percentage as e.g. (0, 0.4) that the augmentation scales the images from 60 % - 140 %, which it should. Could you double check that this is done now?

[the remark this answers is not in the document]
ravinkohli: Hey Lucas, I checked and it is that whatever number we enter, its the percentage of zooming out(if greater than 1) or zooming in(if less than 1). So if we want something like 60-140%, we could have 1 - offset and 1+offset, or we could fix the offset and learn what to use as the centre point. Let me know what you think is more appropriate.

PR #22, opened 2020-10-26
Added backend and Trainer

franchuterivera:
This PR enables to train a pipeline and adds a backend object to it, mainly for logging. 

We will need this backend to store the files for ensemble, but this can be done better when we are coding this later item.


autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -0,0 +1,328 @@
...
+                    key,
+                )
+                for sub_key, sub_value in sorted(value[last_epoch].items()):
+                    string += "\t\t{}: {}\n".format(
+                        sub_key,
+                        sub_value,
+                    )
+            else:
+                string += "\t{}: {}\n".format(
+                    key if 'time' in key else key,
```
ravinkohli, line 145: maybe a typo?

ravinkohli (changes requested):
So the code here looks fine. Is the tensorboard_log [here](https://github.com/franchuterivera/Auto-PyTorch_refactor/blob/e6e0dde9a6c766193c859fb64d11ec247715fd53/autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py#L390) left blank for later? Also, is there a way to get the trained model? I think it could be useful to print the pytorch model and return it after training.

franchuterivera:
Hi @ravinkohli I have addressed your concerns in the PR. I do see travis examples failing due to a missing file, I am doing a push to fix this too. I will text you once it is done

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -0,0 +1,351 @@
...
+    def __init__(self,
+                 budget_type: str,
+                 max_value: int,
+                 ):
+        """
+        An object for tracking when to stop the network training.
+        It handles epoch based criteria as well as training based criteria.
+
+
+        # Maybe it is better to use pynisher? Remember the whole everything gets
```
LMZimmer, line 24: I would prefer to use pynisher, Matthias and I have ownership of that now

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -0,0 +1,351 @@
...
+        copied when forking memory, so it makes sense to ask... but also,
+        how to we regulate memory from GPU and CPU?
```
LMZimmer, line 26: I dont think it is necessary to regulate GPU/CPU memory or do you have a case where it would be? Typically you use all available cores / GPUs and those should be given by the Optimizer/Scheduler

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -0,0 +1,351 @@
...
+        """
+        self.start_time = time.time()
+        self.budget_type = budget_type
+        self.max_value = max_value
```
LMZimmer, line 30: Maybe it is better to have a max_epoch and max_time, both with large defaults? Then we could also consider combined limits like "either 50 epochs or 1h runtime"

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -0,0 +1,351 @@
...
+        return False
+
+    def is_max_time_reached(self) -> bool:
+        elapsed_time = time.time() - self.start_time
+        if self.budget_type == 'runtime' and elapsed_time > self.max_value:
+            return True
+        return False
+
+
+class RunSummary(object):
```
LMZimmer, line 44: Maybe add a description that this is done epoch wise

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -0,0 +1,351 @@
...
+
+    def register_model_statistics(self,
+                                  total_parameter_count: float,
+                                  trainable_parameter_count: float,
+                                  ) -> None:
+        """
+        Utility to register global statistic about the model or training,
+        that are agnostic to epochs
+        """
+        self.total_parameter_count = total_parameter_count
```
LMZimmer, line 103: Is this redundant with the init?

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -0,0 +1,351 @@
...
+
+class BaseTrainerComponent(autoPyTorchTrainingComponent):
+
+    def prepare(
+        self,
+        metrics: List[Any],
+        model: torch.nn.Module,
+        criterion: torch.nn.Module,
+        budget_tracker: BudgetTracker,
+        optimizer: torch.nn.Module,
```
LMZimmer, line 161: I thought that torch.optim.optimizer wasnt an nn.Module (https://pytorch.org/docs/stable/_modules/torch/optim/optimizer.html#Optimizer) or am I wrong?

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -0,0 +1,351 @@
...
+    def on_epoch_end(self, X: Dict[str, Any], epoch: int) -> bool:
+        """
+        Optional place holder for AutoPytorch Extensions.
+        An user can define what happens on every epoch start or every epoch end.
+        If returns True, the training is stopped
+
+        """
+        return False
+
+    def train(self, train_loader: torch.utils.data.DataLoader, epoch: int
```
LMZimmer, line 207: Maybe a better name would be train_epoch or something indicating that this is one epoch

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -0,0 +1,351 @@
...
+            Dict[str, float]: scores for each desired metric
+        '''
+
+        loss_sum = 0.0
+        N = 0
+        self.model.train()
+        outputs_data = list()
+        targets_data = list()
+
+        for step, (data, targets) in enumerate(train_loader):
```
LMZimmer, line 227: This could be outsourced to an additional method `make_step()` such that we could train for a certain number of steps instead of epochs

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -0,0 +1,351 @@
...
+            N += batch_size
+
+            if self.writer:
+                self.writer.add_scalar(
+                    'Train/loss',
+                    loss.item(),
+                    epoch * len(train_loader) + step,
+                )
+
+        return loss_sum / N, self.compute_metrics(outputs_data, targets_data)
```
LMZimmer, line 263: I think we should add an option to not evaluate train metrics for better anytime performance

autoPyTorch/pipeline/components/training/trainer/MixUpTrainer.py
```diff
@@ -0,0 +1,69 @@
...
+        return {
+            'shortname': 'MixUpTrainer',
+            'name': 'MixUp Regularized Trainer',
+        }
+
+    @staticmethod
+    def get_hyperparameter_search_space(dataset_properties: typing.Optional[typing.Dict] = None
+                                        ) -> ConfigurationSpace:
+        alpha = UniformFloatHyperparameter(
+            "alpha", 0, 1, default_value=0.5)
```
LMZimmer, line 66: I think the default should be less aggressive, probably 0.2

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -0,0 +1,351 @@
...
+            X (np.ndarray): The batch training features
+            y (np.ndarray): The batch training labels
+
+        Returns:
+            np.ndarray: that processes data
+            Dict[str, np.ndarray]: arguments to the criterion function
+        """
+        raise NotImplementedError()
+
+    def criterion_preparation(self, y_a: np.ndarray, y_b: np.ndarray = None, lam: float = 1.0
```
LMZimmer, line 337: I assume y_a and y_b are used for mixup. Maybe it would be cleaner to just have y here and in the StandardTrainer and have the MixupTrainer override this method

LMZimmer (changes requested):
Nice work! I left some suggestions

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/utils.py
```diff
@@ -21,9 +21,10 @@ def get_tabular_preprocessers(X: Dict[str, Any]) -> Dict[str, List[BaseEstimator
     for key, value in X.items():
         if isinstance(value, dict):
             if 'numerical' or 'categorical' in value.keys():
```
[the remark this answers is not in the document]
franchuterivera: Yes, I have removed this extra line!

autoPyTorch/pipeline/components/setup/network/base_network.py
```diff
@@ -152,6 +152,25 @@ def to(self, device: Optional[torch.device] = None) -> None:
         else:
             self.network = self.network.to(self.device)
 
+    def predict(self, loader: torch.utils.data.DataLoader) -> torch.Tensor:
+        """
+        Performs batched prediction given a loader object
+        """
+        assert self.network is not None
```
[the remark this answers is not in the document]
franchuterivera: This is exclusive because because of mypy, but yes, user should not see it

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
franchuterivera: Understood. I have added pynisher support both for memory and for runtime

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
franchuterivera: Yeah, I agree the scheduler should manage. This was just a reminder that the scheduler has to be gpu aware, because we don't want to run out of memory in the GPU because we over scheduled. This was a reminder that pynisher cannot control this today nor smac is aware of this (yet).

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
franchuterivera: Added support for this! Following is now possible:

- epoch only support

- runtime only support

- the first budget that is exhausted be it epoch/time

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
franchuterivera: thanks for the suggestion I have renamed this procedure!

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
franchuterivera: I have added more documentation to this class

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
franchuterivera: Yes! I have removed  this

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
franchuterivera: Thanks for catching this, so much for static checking from mypy :D

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
franchuterivera: thanks for the suggestion I have created this function

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
franchuterivera: Done in the new push!

autoPyTorch/pipeline/components/training/trainer/MixUpTrainer.py
franchuterivera: Thanks, I have made this change. I have personally not YET seen this giving really good results, but we will evaluate it once the framework is ready

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
franchuterivera: This is the one comment I cannot address. Mypy does not allow this redefinition as we need to comply with the parent-child implementation. I used to have it as you proposed (because autopytorch originally had it like this) But with mypy, we need to comply with the parent class interface

LMZimmer: Okay, then let's just go with this. This allows us to introduce weights via y_b which is hopefully enough for most cases

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
```diff
@@ -0,0 +1,479 @@
...
+        Returns:
+            ConfigurationSpace: the configuration space of the hyper-parameters of the
+                 chosen component
+        """
+        cs = ConfigurationSpace()
+
+        if dataset_properties is None:
+            dataset_properties = {}
+
+        # Compile a list of legal preprocessors for this problem
```
LMZimmer, line 104: Legal trainers?

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
```diff
@@ -0,0 +1,479 @@
...
+        if X["torch_num_threads"] > 0:
+            torch.set_num_threads(X["torch_num_threads"])
+
+        budget_tracker = BudgetTracker(
+            budget_type=X['budget_type'],
+            max_runtime=X['runtime'] if 'runtime' in X else None,
+            max_epochs=X['epochs'] if 'epochs' in X else None,
+        )
+
+        self.choice.prepare(
```
LMZimmer, line 228: Maybe we could swap `self.choice` for something more descriptive like `self.trainer_choice`

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
```diff
@@ -0,0 +1,479 @@
...
+            model=X['network'],
+            metrics=[m() for m in get_metrics(X['dataset_properties'])],
```
LMZimmer, line 230: Are all metrics determined by the dataset properties? I would imaging that the user also wants to specify additional metrics, e.g. log_loss, accuracy and balanced accuracy

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
```diff
@@ -0,0 +1,479 @@
...
+            criterion=get_loss_instance(X['dataset_properties']),
```
LMZimmer, line 231: Same here, the user possibly wants to specify a custom loss

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
```diff
@@ -0,0 +1,479 @@
...
+                epoch=epoch,
+            )
+
+            val_loss, val_metrics, test_loss, test_metrics = None, {}, None, {}
+            if self.eval_valid_each_epoch(X):
+                val_loss, val_metrics = self.choice.evaluate(X['val_data_loader'], epoch)
+                if 'test_data_loader' in X and X['test_data_loader']:
+                    test_loss, test_metrics = self.choice.evaluate(X['test_data_loader'], epoch)
+
+            # TODO: does it make sense to call all schedulers here?
```
LMZimmer, line 266: You could move the scheduler call to the (Base)Trainer, then we can account for calling them step-wise

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
```diff
@@ -0,0 +1,479 @@
...
+        Args:
+            X (Dict[str, Any]): Dictionary with fitted parameters. It is a message passing
+                mechanism, in which during a transform, a components adds relevant information
+                so that further stages can be properly fitted
+
+        Returns:
+            bool: If true, training should be stopped
+        """
+        assert self.run_summary is not None
+        bad_epochs = self.run_summary.get_best_epoch() - self.run_summary.get_last_epoch()
```
LMZimmer, line 342: Maybe `epochs_since_best` or similar is more fitting

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
```diff
@@ -0,0 +1,479 @@
...
+                so that further stages can be properly fitted
+
+        Returns:
+            bool: if True, the model is evaluated in every epoch
+
+        """
+        if 'early_stopping' in X and X['early_stopping']:
+            return True
+
+        # We need to know if we should reduce the rate based on val loss
```
LMZimmer, line 365: Not strictly, that could also be done when evaluating every e.g. 5 epochs. It would be nice if we could support that

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
```diff
@@ -0,0 +1,479 @@
...
+        Returns the device to do torch operations
+
+        Args:
+            X (Dict[str, Any]): A fit dictionary to control how the pipeline
+                is fitted
+
+        Returns:
+            torch.device: the device in which to compute operations. Cuda/cpu
+        """
+        if not torch.cuda.is_available():
```
LMZimmer, line 451: Could we add a logging info here that CPU or GPU is chosen?

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
```diff
@@ -0,0 +1,381 @@
...
+                string += "\t{}: {}\n".format(
+                    key,
+                    value[last_epoch],
+                )
+        string += '=' * 40
+        return string
+
+
+class BaseTrainerComponent(autoPyTorchTrainingComponent):
+
```
LMZimmer, line 162: A short docstring might be useful here to state that Trainers have to inherit from this class and that `TrainerChoice` will build on top of this

LMZimmer (changes requested):
Thank you for the fixes, I added just a few small remarks 👍

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
franchuterivera: Thanks!, I have fixed this

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
franchuterivera: This we cannot do, because it is the choice framework for dynamically generating the configuration space.

There are choices and there are components. The self.choice is the product of selecting within standard and mixed training

autoPyTorch/pipeline/components/training/trainer/base_trainer.py
franchuterivera: I will add this in the next push!

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
franchuterivera: I will add this in next push!

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
franchuterivera: I will add this in the next push!

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
franchuterivera: Thanks, I think Ravin added support for this, so I missed to integrate this. I will add it now.

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
franchuterivera: Thanks, I think Ravin added support for this, so I missed to integrate this. I will add it now.

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
franchuterivera: yes of course

autoPyTorch/pipeline/components/training/trainer/base_trainer_choice.py
franchuterivera: I can just code something quick to support this, but I will rather add proper checking and support for this, and hence I created https://github.com/LMZimmer/Auto-PyTorch_refactor/issues/24

LMZimmer (approved):

ravinkohli (approved):

PR #29, opened 2020-11-03
Added utitlity to get requirements from each component in the pipeline

### LINKED ISSUE #16: Update the check_requirements code to return the requirements
What is available today:

There is a utility to check that every component has what it requires i the fit dictionary.

What we need:
This utility has to be updated so that it returns the required keys in the fit dictionary so that the API team knows what we expect from the components. The end goal is to have a utility called get_fit_requirements at the base pipeline level, which returns all requirements to fir a given pipeline.

ravinkohli:
Added fit requirements attribute to each component and made utility method in the pipeline to retrieve them

Closes #16

### DIFF
--- a/autoPyTorch/pipeline/components/base_choice.py
+++ b/autoPyTorch/pipeline/components/base_choice.py
@@ -9,6 +9,7 @@ import numpy as np
 from sklearn.utils import check_random_state
 
 from autoPyTorch.pipeline.components.base_component import autoPyTorchComponent
+from autoPyTorch.utils.common import FitRequirement
 
 
 class autoPyTorchChoice(object):
@@ -46,10 +47,15 @@ class autoPyTorchChoice(object):
         # Since the pipeline will initialize the hyperparameters, it is not
         # necessary to do this upon the construction of this object
         # self.set_hyperparameters(self.configuration)
-        self.choice = None
+        self.choice: Optional[autoPyTorchComponent] = None
 
-    def get_fit_requirements(self):
-        return self.choice.get_fit_requirements()
+    def get_fit_requirements(self) -> Optional[List[FitRequirement]]:
+        if self.choice is not None:
+            return self.choice.get_fit_requirements()
+        else:
+            raise AttributeError("Expected choice attribute to be autoPyTorchComponent"
+                                 " but got None, to get fit requirements for {}, "
+                                 "call get_fit_requirements of the component".format(self.__class__.__name__))
 
     def get_components(cls: 'autoPyTorchChoice') -> Dict[str, autoPyTorchComponent]:
         """Returns and ordered dict with the components available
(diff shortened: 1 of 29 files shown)


autoPyTorch/pipeline/components/base_choice.py
```diff
@@ -175,19 +184,17 @@ def get_hyperparameter_search_space(
         """
         raise NotImplementedError()
 
-    def fit(self, X: np.ndarray, y: np.ndarray, **kwargs: Any) -> autoPyTorchComponent:
+    def fit(self, X: Dict[str, Any], y: Any) -> autoPyTorchComponent:
         """Handy method to check if a component is fitted
 
         Args:
-            X (np.ndarray): the input features
+            X (np.ndarray): the fit
```
LMZimmer, line 191: This is not quite clear

autoPyTorch/pipeline/components/preprocessing/image_preprocessing/normalise/base_normalizer.py
```diff
@@ -4,9 +4,16 @@
 
 from autoPyTorch.pipeline.components.preprocessing.image_preprocessing.base_image_preprocessor import \
     autoPyTorchImagePreprocessingComponent
+from autoPyTorch.utils.common import FitRequirement
 
 
 class BaseNormalizer(autoPyTorchImagePreprocessingComponent):
+    _fit_requirements = [FitRequirement('channelwise_mean', np.ndarray),
```
LMZimmer, line 11: With this implemented, we can move parts of the `check_requirements` method to the base Component to prevent redundant code, so that the base component goes through `_fit_requirements` and checks if the key is available in X and checks the type. We can do it now or add an issue to remind ourselves

autoPyTorch/pipeline/components/training/data_loader/base_data_loader.py
```diff
@@ -27,8 +28,12 @@ class BaseDataLoaderComponent(autoPyTorchTrainingComponent):
     https://pytorch.org/docs/stable/data.html
 
     """
+    _fit_requirements = [FitRequirement("dataset", str), FitRequirement("root", str),
```
LMZimmer, line 31: Can we have one requirement per line for better visibility?

LMZimmer (changes requested):
Added very few comments, this is already mergeable though

autoPyTorch/pipeline/components/preprocessing/image_preprocessing/normalise/base_normalizer.py
ravinkohli: okay I'll do that now.

autoPyTorch/pipeline/components/training/data_loader/base_data_loader.py
ravinkohli: Sure

LMZimmer (approved):

PR #34, opened 2020-11-06
Default pipeline options and required properties

ravinkohli:
1. Changed fit_requirements to have multiple supported types as required by some components(especially for input data)
2. Added required properties attribute to components to support defining local properties when adding a third party component
3. Added Utility function to get default pipeline options for setting up an experiment
4. Fixed errors in mypy and running examples

### DIFF
--- a/autoPyTorch/pipeline/base_pipeline.py
+++ b/autoPyTorch/pipeline/base_pipeline.py
@@ -366,3 +366,19 @@ class BasePipeline(Pipeline):
             Dict: Additional information about the pipeline
         """
         return self._additional_run_info
+
+    @staticmethod
+    def get_default_pipeline_options() -> Dict[str, Any]:
+        return {
+            'job_id': '1',
+            'working_dir': '/tmp/experiment_1',
+            'device': 'cpu',
+            'budget_type': 'epochs',
+            'epochs': 20,
+            'runtime': 3600,
+            'torch_num_threads': 1,
+            'early_stopping': 10,
+            'use_tensorboard_logger': True,
+            'use_pynisher': False,
+            'metrics_during_training': True
+        }
--- a/autoPyTorch/pipeline/components/base_component.py
+++ b/autoPyTorch/pipeline/components/base_component.py
@@ -96,7 +96,7 @@ class autoPyTorchComponent(BaseEstimator):
 
     def __init__(self) -> None:
         super().__init__()
-        self._fit_requirements: Optional[List[FitRequirement]] = None
+        self._fit_requirements: List[FitRequirement] = list()
 
     @classmethod
     def get_required_properties(cls) -> Optional[List[str]]:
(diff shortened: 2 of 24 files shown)


autoPyTorch/pipeline/base_pipeline.py
```diff
@@ -366,3 +366,19 @@ def get_additional_run_info(self) -> Dict:
...
+            'working_dir': '/tmp/experiment_1',
+            'device': 'cpu',
+            'budget_type': 'epochs',
+            'epochs': 20,
+            'runtime': 3600,
+            'torch_num_threads': 1,
+            'early_stopping': 10,
+            'use_tensorboard_logger': True,
+            'use_pynisher': False,
+            'metrics_during_training': True
```
LMZimmer, line 383: For my information: what is `metrics_during_training` used for?

autoPyTorch/pipeline/components/base_component.py
```diff
@@ -203,14 +219,22 @@ def check_requirements(self, X: Dict[str, Any], y: Any = None) -> None:
...
 
-        for requirement in self._fit_requirements:
-            if requirement.name not in X.keys():
-                raise ValueError("To fit {}, expected fit dictionary to have '{}'"
-                                 " but got \n {}".format(self.__class__.__name__, requirement.name, list(X.keys())))
-            elif not isinstance(X[requirement.name], requirement.type):
-                raise TypeError("Expected {} to be instance of {} got {}".format(requirement.name,
-                                                                                 requirement.type,
-                                                                                 type(X[requirement.name])))
+        if self._fit_requirements:
```
LMZimmer, line 222: Better `if self._fit_requirements is not None`, or even better initialize `self._fit_requirements: Optional[List[FitRequirement]] = list()` in line 95 and delete the if statement

autoPyTorch/pipeline/components/base_component.py
```diff
@@ -203,14 +219,22 @@ def check_requirements(self, X: Dict[str, Any], y: Any = None) -> None:
...
+            for requirement in self._fit_requirements:
+                if requirement.name not in X.keys():
+                    raise ValueError("To fit {}, expected fit dictionary to have '{}'"
+                                     " but got \n {}".format(self.__class__.__name__, requirement.name, list(X.keys())))
+                else:
+                    UNSUPPORTED_TYPE = True
+                    for supported_type in requirement.supported_types:
+                        if isinstance(X[requirement.name], supported_type):
```
LMZimmer, line 230: you can get rid of this for loop and variable assignment like this:

`TYPE_SUPPORTED = isinstance(X[requirement.name], tuple(requirement.supported_types))`

since `isinstance` supports checking multiple types given as a tuple

autoPyTorch/pipeline/components/preprocessing/image_preprocessing/normalise/base_normalizer.py
```diff
@@ -11,9 +11,11 @@ class BaseNormalizer(autoPyTorchImagePreprocessingComponent):
...
     def __init__(self) -> None:
         super(BaseNormalizer, self).__init__()
-        self._fit_requirements = [FitRequirement('channelwise_mean', np.ndarray),
-                                  FitRequirement('channelwise_std', np.ndarray)]
-        self._fit_requirements.extend(super()._fit_requirements)
+        self._fit_requirements = [FitRequirement('channelwise_mean', (np.ndarray,)),
+                                  FitRequirement('channelwise_std', (np.ndarray,))]
+        super_requirements: Optional[List[FitRequirement]] = super().get_fit_requirements()
+        if super_requirements:
+            self._fit_requirements.extend(super_requirements)
```
LMZimmer, line 18: Thinking about this a bit more, it would probably be more elegant to provide an `add_requirement` method in the base component Then all components just have to call add requirement in their init. That way, one does not have to be careful to extend the super requirements when adding a new component.

LMZimmer (changes requested):

autoPyTorch/pipeline/base_pipeline.py
ravinkohli: So, if it is true, then after every epoch the metrics are computed and stored in the summary along with the loss, if it is fale, then only the loss is stored

LMZimmer (approved):

PR #36, opened 2020-11-12
Ensemble builder

franchuterivera:
Initial Commit with an empty file, mainly to test the check-api


autoPyTorch/ensemble/abstract_ensemble.py
```diff
@@ -0,0 +1,78 @@
...
+    ) -> 'AbstractEnsemble':
+        """Fit an ensemble given predictions of base models and targets.
+        Ensemble building maximizes performance (in contrast to
+        hyperparameter optimization)!
+
+        Args:
+            base_models_predictions (np.ndarray):
+                array of shape = [n_base_models, n_data_points, n_targets]
+                n_targets is the number of classes in case of classification,
+                n_targets is 0 or 1 in case of regression
```
ravinkohli, line 27: do you mean its between 0 and 1 in the case of regression?

autoPyTorch/ensemble/ensemble_builder.py
```diff
@@ -0,0 +1,1432 @@
...
+                    os.remove(self.ensemble_memory_file)
+                except:  # noqa E722
+                    pass
+
+                if isinstance(self.ensemble_nbest, numbers.Integral) and self.ensemble_nbest <= 1:
+                    if self.read_at_most == 1:
+                        self.logger.error(
+                            "Memory Exception -- Unable to further reduce the number of ensemble "
+                            "members and can no further limit the number of ensemble members "
+                            "loaded per iteration -- please restart Auto-sklearn with a higher "
```
ravinkohli, line 630: AutoPyTorch, :)

ravinkohli:
Should we wait for the dataset class to be merged and then make changes to this PR before merging it? Otherwise, for me the changes are good.

franchuterivera:
We need the changes in this PR for SMAC. It also fixes the backend

When building an ensemble we use smac output, so we do not care about the data manager. 

Nevertheless, I am reading the datamanager to make any time test performance, but is a 1 line change when the dataset/datamanager is ready

ravinkohli:
> We need the changes in this PR for SMAC. It also fixes the backend
> 
> When building an ensemble we use SMAC output, so we do not care about the data manager.
> 
> Nevertheless, I am reading the data manager to make any time test performance, but is a 1 line change when the dataset/data manager is ready
Okay then, from my side it can be merged. Let's wait for Lucas to review it

autoPyTorch/data/abstract_data_manager.py
```diff
@@ -0,0 +1,34 @@
+import abc
+from typing import Any, Dict, List
+
+import numpy as np
+
+
+class AbstractDataManager():
```
LMZimmer, line 7: Maybe we could add some documentation here?

autoPyTorch/ensemble/ensemble_builder.py
```diff
@@ -0,0 +1,1432 @@
...
+                              (ensemble_nbest, type(ensemble_nbest)))
+
+        self.start_time = 0.0
+        self.model_fn_re = re.compile(MODEL_FN_RE)
+
+        self.last_hash = None  # hash of ensemble training data
+        self.y_true_ensemble = None
+        self.SAVE2DISC = True
+
+        # already read prediction files
```
LMZimmer, line 481: Do we still need this?

autoPyTorch/ensemble/ensemble_builder.py
```diff
@@ -0,0 +1,1432 @@
...
+        time_left: Optional[float] = None,
+        end_at: Optional[float] = None,
+        time_buffer: int = 5,
+        return_predictions: bool = False,
+    ) -> Tuple[
+        List[Dict[str, float]],
+        int,
+        Optional[np.ndarray],
+        Optional[np.ndarray],
+    ]:
```
LMZimmer, line 582: Could we add a docstring here?

autoPyTorch/ensemble/ensemble_builder.py
```diff
@@ -0,0 +1,1432 @@
...
+                        self.ensemble_nbest = int(self.ensemble_nbest / 2)
+                    self.logger.warning("Memory Exception -- restart with "
+                                        "less ensemble_nbest: %d" % self.ensemble_nbest)
+                    return [], self.ensemble_nbest, None, None
+            else:
+                return safe_ensemble_script.result
+
+        return [], self.ensemble_nbest, None, None
+
+    def main(
```
LMZimmer, line 656: Could we maybe add a short docstring?

autoPyTorch/pipeline/base_pipeline.py
```diff
@@ -129,6 +129,8 @@ def predict(self, X: np.ndarray, batch_size: Optional[int] = None
         """
 
         # Pre-process X
+        if batch_size is None:
+            batch_size = X.shape[0]
```
LMZimmer, line 133: Could we add an info that batch size is set automatically here?

LMZimmer (changes requested):
Thanks for the work! Some points:

1. Looks like some files are geting through the .gitignore in `test/test_ensemble/.autoPyTorch`

2. We might want to provide a function to get the ensemble models/directories/pipelines since this is something many users wanted. Let's either provide that now or create an issue to keep this in mind

BryanLaura (approved):

BryanLaura (approved):

autoPyTorch/data/abstract_data_manager.py
franchuterivera: This was a WA because we didn't have the dataset class! I am going to rebase and remove this which is no longer needed

autoPyTorch/ensemble/ensemble_builder.py
franchuterivera: Yes and I have update the comment as to why:
```
  # already read prediction files
         # We read in back this object to give the ensemble the possibility to have memory
         # Every ensemble task is sent to dask as a function, that cannot take un-picklable
         # objects as attributes. For this reason, we dump to disk the stage of the past
         # ensemble iterations to kick-start the ensembling process

```

autoPyTorch/datasets/base_dataset.py
```diff
@@ -57,19 +117,27 @@ def _get_indices(self) -> np.ndarray:
...
-        return [(Subset(self, split[0]), Subset(self, split[1])) for split in splits]
+                                cross_val_type: CrossValTypes,
+                                num_splits: int) -> None:
+        """
+        This function creates the cross validation split for the given task.
+
+        It is done once per dataset to have comparable results among pipelines
+        """
+        # Create just the split once
+        if self.splits is None:
```
LMZimmer, line 128: Maybe you could add an `else:` with a warning/error when this is called after already creating a split?

autoPyTorch/datasets/base_dataset.py
```diff
@@ -81,16 +149,76 @@ def create_val_split(self,
...
-            train, val = self.holdout_validators[holdout_val_type](val_share, self._get_indices(), **kwargs)
+            train, val = self.holdout_validators[holdout_val_type.name](val_share, self._get_indices(), **kwargs)
             return Subset(self, train), Subset(self, val)
         else:
             if self.val_tensors is None:
                 raise ValueError('Please specify `val_share` or initialize with a validation dataset.')
             val = BaseDataset(self.val_tensors)
             return self, val
+
+    def get_dataset_for_training(self, split: int) -> Tuple[Dataset, Dataset]:
```
LMZimmer, line 166: Maybe something like `split_id` is more descriptive (also since this occurs later in the pipeline as a config option)

autoPyTorch/datasets/base_dataset.py
```diff
@@ -81,16 +149,76 @@ def create_val_split(self,
...
+                holdout_val_type=self.resampling_strategy,
+                val_share=val_share,
+            )
+        elif isinstance(self.resampling_strategy, CrossValTypes):
+            num_splits = DEFAULT_RESAMPLING_PARAMETERS[self.resampling_strategy].get(
+                'num_splits', None),
+            if self.resampling_strategy_args is not None:
+                num_splits = self.resampling_strategy_args.get('num_splits', num_splits)
+            # Create the split if it was not created before
+            self.create_cross_val_splits(
```
LMZimmer, line 195: If you create the warning as mentioned above, this could have an `if self.splits is None:`

autoPyTorch/pipeline/components/training/data_loader/base_data_loader.py
```diff
@@ -49,8 +47,9 @@ def __init__(self, batch_size: int = 64) -> None:
         self.add_fit_requirements([
             FitRequirement("dataset", (str,), user_defined=True, dataset_property=True),
             FitRequirement("root", (str,), user_defined=True, dataset_property=True),
-            FitRequirement("X_train", (np.ndarray,), user_defined=True, dataset_property=True),
+            FitRequirement("split", (int,), user_defined=True, dataset_property=True),
```
LMZimmer, line 50: Same as above, `split_id` might be more expressive

examples/example_ensemble_classification.py
```diff
@@ -0,0 +1,227 @@
...
+        'torch_num_threads': 1,
+        'early_stopping': 20,
+        'use_tensorboard_logger': True,
+        'use_pynisher': False,
+        'memory_limit': 4096,
+        'metrics_during_training': True,
+        'seed': 0,
+        'budget_type': 'epochs',
+        'epochs': 10.0,
+        'id': 0,
```
LMZimmer, line 103: `id` is a Python built-in, so I'd prefer another name (I think autosklearn also uses this)

LMZimmer (changes requested):
Just very minor comments. Also, there still are a bunch of unnecessary files in `test/test_ensemble/data/.autoPyTorch` that should be removed (just delete them locally, commit and then maybe add `test/test_ensemble/data/.autoPyTorch` to the `.gitignore`)

autoPyTorch/datasets/base_dataset.py
franchuterivera: I have clarified why I do expect this function to be called multiple times via a comment:
```
      # Create just the split once
         # This is gonna be called multiple times, because the current dataset
         # is being used for multiple pipelines. That is, to be efficient with memory
         # we dump the dataset to memory and read it on a need basis. So this function
         # should be robust against multiple calls, and it does so by remembering the splits

```

autoPyTorch/datasets/base_dataset.py
franchuterivera: Ah, ok cool... processing this comment serially... so I will do that!

examples/example_ensemble_classification.py
franchuterivera: I think this is not needed and I have removed it.

autoPyTorch/pipeline/components/training/data_loader/base_data_loader.py
franchuterivera: Done!

PR #40, opened 2020-11-18
Reimplementation of autopytorchmetric

### LINKED ISSUE #21: Add a check that makes sure each supported metric works
Add a checker to make sure we create an instance of each metric object and make sure it is working. For example:

https://github.com/LMZimmer/Auto-PyTorch_refactor/blob/acc7eebe9641a995933c9243c1308b7e23b7ab08/autoPyTorch/pipeline/components/training/metrics/Accuracy.py#L22

Assumes that pytorch lighting has a pytorchlighting.classification.Accuracy object but it should be pytorch_lightning.metrics.classification.Accuracy

ravinkohli:
Changes made:
1. autoPyTorchMetric is now a wrapper on sklearn metrics
2. Added _ProbMetric, _ThresholdMetric and _PredictMetric based on [autosklearn](https://github.com/automl/auto-sklearn/blob/master/autosklearn/metrics/__init__.py)
3. Made other utility functions in autosklearn metrics compatible with get_metrics and get_supported_metrics
4. Changed compute metric function in BaseTrainerComponent and added task_type in prepare which is needed by compute_metric
5. Added unittests for the same

Reasons to switch from PyTorch lightning to sklearn:
1. PyTorch lightning in their recent update removed class metrics used in the old implementation. 
2. sklearn functionals perform faster than pytorch lightning functionals even for tensors (tested locally on my laptop)
3. Elegant implementation in autosklearn that handles various prediction and target types

Also, Resolves #21


LMZimmer (approved):

LMZimmer:
We may have to make some small changes with the dataset PR but for now this can be merged as is 👍

PR #43, opened 2020-11-30
Small fix to ensure that get_requirements only returns user required inputs

LMZimmer:
There were keys such as `optimizer` in the requirements as returned by `get_requirements` that are generated by the pipeline itself. This introduces an attribute `user_defined` to the `FitRequirement` class such that `get_requirements` can filter by these.


ravinkohli (approved):
I think this is a good solution for the dataset properties.

ravinkohli (commented):
Should we have dataset_property to be a flag that is used to create the dataset_property dictionary that currently just expects 'numerical_columns', 'categorical_columns', 'task_type', 'output_type'? We won't have to pass X_train around in two dictionaries, i.e, fit_dictionary and dataset_properties?

PR #46, opened 2020-12-04
Dataset properties

ravinkohli:
Changes made:
1.  changed constants to comply with sklearn type_of_target
2.  changed dataset_property flag for few keys like X_train, train_indices as these will be added by the AutoML class after performing splits on the dataset. 
3.  added interpret_dataset_properties to tabular dataset
4.  added get_dataset_properties in base dataset


autoPyTorch/constants.py
```diff
@@ -21,20 +23,23 @@
      'image_regression': IMAGE_REGRESSION}
 
 BINARY = 10
```
LMZimmer, line 25: Maybe we could add a comment here, saying that naming is the same as sklearns type_of_target (https://scikit-learn.org/stable/modules/generated/sklearn.utils.multiclass.type_of_target.html)

autoPyTorch/datasets/base_dataset.py
```diff
@@ -231,3 +244,9 @@ def replace_data(self, X_train: BASE_DATASET_INPUT, X_test: Optional[BASE_DATASE
         if X_test is not None and self.test_tensors is not None:
```
LMZimmer, line 244: I think this if statement should only be `if X_test is not None`, that is the behavior I would expect no?

autoPyTorch/datasets/tabular_dataset.py
```diff
@@ -56,21 +57,25 @@ def __init__(self, X: Any, Y: Any,
             # the below function will simply return Pandas DataFrame.
             Y = check_array(Y, ensure_2d=False)
 
+        self.categorical_columns, self.numerical_columns, self.categories, self.num_features, self.num_classes = \
+            self.interpret_dataset_properties(X, Y)
```
LMZimmer, line 61: Maybe `infer_dataset_properties` would be more accurate

autoPyTorch/pipeline/components/preprocessing/base_preprocessing.py
```diff
@@ -23,7 +23,7 @@ def __init__(self) -> None:
         self.add_fit_requirements([
             FitRequirement('X_train',
                            (np.ndarray, pd.DataFrame, csr_matrix),
-                           user_defined=True, dataset_property=True)])
+                           user_defined=True, dataset_property=False)])
```
LMZimmer, line 26: Just double-checking here: setting this to False would mean that X_train shows as possible user input, is that what you need? Because if not then I'd leave it as True, then we can enforce that it only arrives at `fit` via the dataset class

autoPyTorch/pipeline/components/setup/early_preprocessor/EarlyPreprocessing.py
```diff
@@ -20,8 +20,9 @@ def __init__(self, random_state: Optional[np.random.RandomState] = None) -> None
         self.random_state = random_state
         self.add_fit_requirements([
             FitRequirement('is_small_preprocess', (bool,), user_defined=True, dataset_property=True),
-            FitRequirement('X_train', (np.ndarray, pd.DataFrame, csr_matrix), user_defined=True, dataset_property=True),
-            FitRequirement('train_indices', (List,), user_defined=True, dataset_property=True)])
+            FitRequirement('X_train', (np.ndarray, pd.DataFrame, csr_matrix), user_defined=True,
```
LMZimmer, line 23: Same question as above

autoPyTorch/pipeline/components/training/data_loader/base_data_loader.py
```diff
@@ -45,10 +45,8 @@ def __init__(self, batch_size: int = 64) -> None:
 
         # Define fit requirements
         self.add_fit_requirements([
-            FitRequirement("dataset", (str,), user_defined=True, dataset_property=True),
-            FitRequirement("root", (str,), user_defined=True, dataset_property=True),
-            FitRequirement("split_id", (int,), user_defined=True, dataset_property=True),
-            FitRequirement("train_indices", (List[int],), user_defined=True, dataset_property=True),
+            FitRequirement("split_id", (int,), user_defined=True, dataset_property=False),
```
LMZimmer, line 48: Again

autoPyTorch/pipeline/tabular_regression.py
```diff
@@ -0,0 +1,208 @@
...
+    def score(self, X: np.ndarray, y: np.ndarray, batch_size: Optional[int] = None) -> np.ndarray:
+        """score.
+
+                Args:
+                    X (np.ndarray): input to the pipeline, from which to guess targets
+                    batch_size (Optional[int]): batch_size controls whether the pipeline
+                        will be called on small chunks of the data. Useful when calling the
+                        predict method on the whole array X results in a MemoryError.
+                Returns:
+                    np.ndarray: coefficient of determination R^2 of the prediction
```
LMZimmer, line 109: We probably should add support for other metrics here (can be a new issue, if you need this to be done quickly)

LMZimmer (commented):
Just some small comments 👍

autoPyTorch/pipeline/components/preprocessing/base_preprocessing.py
ravinkohli: my idea was that we set dataset property to be true only for those requirements that need to be in the dataset properties dictionary. So, the requirements which have this flag(dataset_property) true are those that the method infer_dataset_properties will calculate and then return as a dict which can be added to the fit requirements either in the AutoML class or by the API. Also, X_train would either be loaded from the datamanager directly at early preprocessing, if it is small enough or in the data loader class, in my opinion, it need not be passed in the fit dictionary from the start. What do you think about this?

autoPyTorch/pipeline/components/training/data_loader/base_data_loader.py
ravinkohli: Also for split id, they will again be calculated based on the fold that is being executed in case of cross validation or 0 in case of holdout. As a single dataset could be used for multiple runs of the pipeline and therefore, it should only contain the splits calculated at the beginning and the split id should be handled in the API or the AutoML class which can give different values for different runs of the pipeline in the same experiment

autoPyTorch/pipeline/components/preprocessing/base_preprocessing.py
LMZimmer: I think that would be a clean solution (removing `X_train` from the requirements), also because in cross-validation these will differ anyways. One could also just have `X` and `split_id` as input.

About the flags, I think a `dataset_property=True` indicates that this is a quantity provided by the dataset (not necessarily in the `dataset_properties` dict, we might throw that argument out later anyways and just use its keys). 

`user_defined=True` indicates this is an input to be given when calling `pipeline.fit`. Conversely, `user_defined=False` indicates this is a `fit_requirement` produced by the pipeline itself (e.g. the initialized optimizer).

autoPyTorch/pipeline/components/training/data_loader/base_data_loader.py
LMZimmer: Yes, I see the comment above 👍

autoPyTorch/pipeline/components/preprocessing/base_preprocessing.py
ravinkohli: I think yes, we could just have the split id in the fit dictionary. We don't need X also there as it is only needed in the early preprocessing node. So, if the dataset is small enough to be preprocessed we can get the train and val dataset directly at the preprocessing node with the backend and split id, and add this to the fit dictionary after preprocessing, then at the data loader stage if the dataset was preprocessed, we use the data present in the fit dictionary or else, we use the backend and split id to get the data there.

autoPyTorch/pipeline/tabular_regression.py
ravinkohli: Currently, I just added this for the optimizer as there were no regression tasks. I'll make an issue but I don't need it to be done quickly

autoPyTorch/datasets/base_dataset.py
ravinkohli: Hey, actually this X_test is coming from the fit dictionary and the test_tensors are coming from the base dataset. Basically, the X_test would be the preprocessed data(if it's small enough) coming from the early preprocessing node. So this replace data function would put the preprocessed X_train and X_test from the fit dictionary into the data loader else it would use from the backend.

LMZimmer: i see, then this is resolved

Issue #49, opened 2020-12-09
Improve the way we decide if all data can fit in memory


franchuterivera:
The following attribute decides if all data can fit in memory and hence, we will load it and pre-process it before fit:

https://github.com/LMZimmer/Auto-PyTorch_refactor/blob/ac7a9ce35e87a428caca2ac108b362a54d3b8f3a/autoPyTorch/datasets/base_dataset.py#L76

If such is not the case, we load it in batches through the data loader.

We can proactively use the [memory_usage](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.memory_usage.html) feature from pandas to estimate the consumption of the full dataset. 

To decide if a dataset fits in the given memory, one can use the memory limit specified in the fit dictionary to come up with how much memory we can use (or default to psutils memory query to see the total memory in the system). The burden of fitting a pipeline (regarding to memory) is a factor of the dataset size. We can create a heuristic to say how much memory a pipeline cost, but in reality, this is hard to measure. This feature of pre-processing the whole data, to my personal criteria, should be done for a really small dataset. So if a dataset consumes say 'half' or a 'third' of the total virtual memory, we should turn this on.

To decide on whether we do it when the dataset consumes half, a third, a fourth can be empirically measured. Let me know if you have a better idea!

PR #50, opened 2020-12-09
Automatically infer is_small_dataset

LMZimmer:
Adresses #49

### DIFF
--- a/autoPyTorch/datasets/base_dataset.py
+++ b/autoPyTorch/datasets/base_dataset.py
@@ -1,7 +1,7 @@
 import sys
 import warnings
 from abc import ABCMeta
-from typing import Any, Dict, Iterable, List, Optional, Tuple, Union, cast
+from typing import Any, Dict, List, Optional, Tuple, Union, cast
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


autoPyTorch/datasets/base_dataset.py
```diff
@@ -82,6 +97,17 @@ def __init__(
         # or for augmentation
         self.transform = transforms
 
+    @staticmethod
+    def is_small_dataset(data_tensors: Iterable[BASE_DATASET_INPUT]) -> bool:
+        """
+        Determine wether the whole dataset can be preprocessed in memory.
+        :param data_tensors: An iterable holding tensors used to infer the dataset size
+        """
+        available_memory_bytes = psutil.virtual_memory().available
```
franchuterivera: Do you think we can take into account the fit dictionary memory usage? For instance, if in a cluster you have 100 Gb of memory, but you are only allowed few Gb, psutil will always return 100Gb, so any dataset will be small.

autoPyTorch/datasets/base_dataset.py
```diff
@@ -40,6 +45,17 @@ def type_check(train_tensors: BASE_DATASET_INPUT, val_tensors: Optional[BASE_DAT
             check_valid_data(val_tensors[i])
 
 
+def get_tensor_size(data_tensor: BASE_DATASET_INPUT) -> int:
```
franchuterivera, line 48: Do you think we can add a small test to make sure this flag is_small_preprocess is asserted only when it makes sense?

For example, this variable is true if a np.array of 10 elements is provided but not true if a np.array of 30000 elements is provided. Just to make sure we don't break this in the future.

LMZimmer: Will do that tomorrow, I think I'll have to create a new test module for the `BaseDataset` for that

autoPyTorch/datasets/base_dataset.py
LMZimmer: Mhh true, I guess the API could pass that to the dataset and call the method again.

Another way might be to just `try - except` do something with the dataset and catch a memory error, maybe I'll add both

ravinkohli (approved):
For me, It can be merged after tests are added

PR #51, opened 2020-12-10
New networks

bastiscode:
PR includes ShapedResnet backbone for tabular data and Resnet backbone for image data. More coming soon.

### DIFF
--- a/autoPyTorch/pipeline/components/setup/network/utils/common.py
+++ b/autoPyTorch/pipeline/components/setup/network/utils/common.py
@@ -1,0 +1,53 @@
+from typing import Tuple
+
+import torch.nn as nn
+import math
+
+
+def initialize_weights(module: nn.Module) -> None:
+    if isinstance(module, nn.Conv2d):
+        n = module.kernel_size[0] * module.kernel_size[1] * module.out_channels
+        module.weight.data.normal_(0, math.sqrt(2. / n))
+        # nn.init.kaiming_normal(module.weight.data, mode='fan_out')
+    elif isinstance(module, nn.BatchNorm2d):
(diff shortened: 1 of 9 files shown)


LMZimmer:
I have a general question about the backbone part. Do you think it is possible to share code between `image.py`, `tabular.py` and `time_series.py` - for instance, putting conv/dense layers into a `layers.py`, or even have a conv net that can be used by all of them?

bastiscode:
I also thought about that. I think for stuff like attention layers which could be possibly added later a `layers.py` makes sense. If one wants to include more NAS-like capabilities, I think you could implement searchable dense and conv layers or other module blocks that can be used in backbones across tasks. Then you could also add more general backbones like NASImageBackbone for example.

But as of now I think all the implemented networks have a rather specific architecture, and sharing code between them is not something I would do.

LMZimmer:
Okay, then I will add that to the todo/future work 👍

autoPyTorch/pipeline/components/setup/network/utils/common.py
```diff
@@ -0,0 +1,53 @@
+from typing import Tuple
+
+import torch.nn as nn
+import math
+
+
+def initialize_weights(module: nn.Module) -> None:
```
LMZimmer, line 7: Don't we already have an initializer in https://github.com/LMZimmer/Auto-PyTorch_refactor/blob/development/autoPyTorch/pipeline/components/setup/network_initializer/base_network_initializer.py ?

bastiscode: Yes, I will remove it from the network.

PR #54, opened 2020-12-10
SMAC-integration

ravinkohli:
Major changes include

1. Added autopytorch/evaluation where 
  1.1. tae.py contains the ExecuteTAFuncWithQueue. 
  1.2. train_evaluater.py contains eval_function closure to be used in ExecuteTAFuncWithQueue. Currently, only tested with holdout and cross validation on tabular tasks.
2. Changed way of creating splits in BaseDataset. now it will initialise splits attribute of the BaseDataset.
3. Made dataset_properties as a key in fit dictionary which contains output of datamanager.get_dataset_properties
4. Added predict_proba function to the pipeline


.idea/Auto-PyTorch_refactor.iml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
LMZimmer: Should these be here?

autoPyTorch/datasets/base_dataset.py
```diff
@@ -68,10 +67,12 @@ def __init__(
         self.shuffle = shuffle
         self.resampling_strategy = resampling_strategy
         self.resampling_strategy_args = resampling_strategy_args
-        self.task_type: Optional[int] = None
+        self.task_type: Optional[str] = None
         self.issparse: bool = issparse(self.train_tensors[0])
-        self.output_type: int = STRING_TO_OUTPUT_TYPES[type_of_target(self.train_tensors[1])]
-
+        self.num_classes = None
+        if len(train_tensors) == 2 and train_tensors[1] is not None:
```
LMZimmer, line 73: What happens with `num_classes` and `output_type` if this is not true?

autoPyTorch/pipeline/components/base_component.py
```diff
@@ -224,18 +224,22 @@ def check_requirements(self, X: Dict[str, Any], y: Any = None) -> None:
 
         if y is not None:
             warnings.warn("Provided y argument, yet only X is required")
+        assert 'dataset_properties' in X.keys(), "Fit dictionary must contain dataset_properties"
 
         for requirement in self._fit_requirements:
-            if requirement.name not in X.keys():
+            check_dict = X
```
LMZimmer: `check_dict = X['dataset_properties'] if requirement.dataset_property else X` might be more readable

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/scaling/StandardScaler.py
```diff
@@ -22,7 +22,7 @@ def __init__(self,
     def fit(self, X: Dict[str, Any], y: Any = None) -> BaseScaler:
 
         self.check_requirements(X, y)
-
+        # TODO: have checking for sparsity done in Dataset
```
LMZimmer: Can we put this in the gdoc / in an issue? Here we might forget it

autoPyTorch/pipeline/components/training/metrics/utils.py
```diff
@@ -83,20 +92,19 @@ def get_metrics(dataset_properties: Dict[str, Any],
...
 
     return metrics
 
 
 def calculate_score(
-    solution: np.ndarray,
-    prediction: np.ndarray,
-    task_type: int,
-    metrics: Iterable[autoPyTorchMetric],
+        solution: np.ndarray,
```
LMZimmer: I'd prefer target or label

autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -0,0 +1,533 @@
...
+
+__all__ = [
+    'AbstractEvaluator',
+    'fit_and_suppress_warnings'
+]
+
+
+# TODO: make dummypipeline or model for autopytorch
+class MyDummyClassifier(DummyClassifier):
+    def __init__(self, config: Configuration,
```
LMZimmer, line 54: Could we write some explaining docstrings in this module?

LMZimmer (commented):
Started to look at the PR, more comments coming tomorrow

autoPyTorch/datasets/base_dataset.py
ravinkohli: That would be the case when we're doing unsupervised learning. currently, I am not taking care of that as I focused on classification and regression. What would you suggest to do? Should I create output constants for unsupervised learning as well? Or for now, we can also raise an error saying that the task is currently not supported.

autoPyTorch/pipeline/components/preprocessing/tabular_preprocessing/scaling/StandardScaler.py
ravinkohli: Actually it has been implemented already so I'll remove this and use the issparse coming from the dataset

autoPyTorch/evaluation/abstract_evaluator.py
ravinkohli: Yes I'll do that

autoPyTorch/utils/pipeline.py
```diff
@@ -62,19 +66,10 @@ def get_dataset_requirements(info: Dict[str, Any],
 
 def _get_regression_dataset_requirements(info: Dict[str, Any], include: Dict[str, List[str]],
```
LMZimmer, line 67: What is `info` here? Probably could use a more descriptive name

autoPyTorch/utils/pipeline.py
```diff
@@ -1,10 +1,13 @@
 # -*- encoding: utf-8 -*-
```
LMZimmer, line 1: I am thinking if the methods in this module would be better in the pipeline classes. E.g.:

- add a `self.task_type` attribute to the pipelines (maybe not even needed)
- the `BasePipeline` already implements `get_hyperparameter_search_space` (= the `get_configspace` function from here), it just needs the include/exclude support
- `get_dataset_requirements` is also already implemented in the `BasePipeline`
- create an util function `get_pipeline(task_type)` that returns the initialized pipeline. on that one can then call one of the above two methods

What do you think?

autoPyTorch/pipeline/base_pipeline.py
```diff
@@ -408,7 +408,7 @@ def get_default_pipeline_options() -> Dict[str, Any]:
             'job_id': '1',
             'device': 'cpu',
             'budget_type': 'epochs',
-            'epochs': 20,
+            'epochs': 1,
```
LMZimmer: I assume this was for testing? We probably want 20 as default

autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -0,0 +1,533 @@
...
+        # TODO: adjust backend for autopytorch backend
+        self.backend = backend
+        self.queue = queue
+
+        # TODO: use autopytorch datasets once PR is passed
+        self.datamanager: BaseDataset = self.backend.load_datamanager()
+        self.include = include
+        self.exclude = exclude
+
+        self.X_train, self.y_train = self.datamanager.train_tensors
```
LMZimmer, line 183: How are `X_val` and `y_val` handled here?

autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -0,0 +1,533 @@
...
+            self.X_test, self.y_test = None, None
+
+        self.metric = metric
+        self.task_type = STRING_TO_TASK_TYPES[self.datamanager.task_type]
+        self.output_type = STRING_TO_OUTPUT_TYPES[self.datamanager.output_type]
+        self.issparse = self.datamanager.issparse
+
+        self.seed = seed
+
+        self.output_y_hat_optimization = output_y_hat_optimization
```
LMZimmer, line 200: what is this variable for?

autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -0,0 +1,533 @@
...
+        # TODO: Check if we need all supported metrics, as in our case even single metric is in a score_dict form
+
+        if isinstance(disable_file_output, (bool, list)):
+            self.disable_file_output = disable_file_output
+        else:
+            raise ValueError('disable_file_output should be either a bool or a list')
+
+        self.model_class: Optional[BaseEstimator] = None
```
LMZimmer: This block seems like it could be shrinked significantly with the `get_pipeline(task_type)` method mentioned in the pipeline/utils below

autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -0,0 +1,533 @@
...
+                default_pipeline_options['runtime'] = self.budget
+                if 'epochs' in default_pipeline_options:
+                    del default_pipeline_options['epochs']
+            elif self.budget_type == 'epochs':
+                default_pipeline_options['epochs'] = self.budget
+                if 'runtime' in default_pipeline_options:
+                    del default_pipeline_options['runtime']
+            default_pipeline_options['budget_type'] = self.budget_type
+        self.logger.debug("Default pipeline options: {}".format(default_pipeline_options))
+        self._init_params.update({**default_pipeline_options})
```
LMZimmer: This would override any parameters that have previously been specified if they also occur in the default options. Maybe `self._init_params = {**default_pipeline_options, **self._init_params}` would be safer

autoPyTorch/evaluation/abstract_evaluator.py
LMZimmer: Maybe `pipeline_class` instead of `model_class` and `DummyClassificationPipeline`/`DummyRegressionPipeline` is more descriptive

autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -0,0 +1,533 @@
...
+    def _get_model(self) -> BaseEstimator:
+        assert self.model_class is not None, "Can't return model, model_class not initialised"
+        if not isinstance(self.configuration, Configuration):
+            model = self.model_class(config=self.configuration,
+                                     random_state=np.random.RandomState(self.seed),
+                                     init_params=self._init_params)
+        else:
+            if self.configuration is not None:
+                self.logger.debug("model_class is : {}".format(self.model_class))
+            model = self.model_class(config=self.configuration,
```
LMZimmer: What is the difference between `configuration` and `init_params` here? Also `model` -> `pipeline` is more descriptive

autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -0,0 +1,533 @@
...
+                                     dataset_properties=self.dataset_properties,
+                                     random_state=np.random.RandomState(self.seed),
+                                     include=self.include,
+                                     exclude=self.exclude,
+                                     init_params=self._init_params)
+        return model
+
+    def _loss(self, y_true: np.ndarray, y_hat: np.ndarray) -> Dict[str, float]:
+        """Auto-sklearn follows a minimization goal, so the make_scorer
```
LMZimmer: Maybe "SMAC follows a minimization goal"?

autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -0,0 +1,533 @@
...
+        score = calculate_score(
+            y_true, y_hat, self.task_type, self.metric)
+
+        err = {metric.name: metric._optimum - score[metric.name] for metric in self.metric
+               if metric.name in score.keys()}
+
+        return err
+
+    def finish_up(self, loss: Dict[str, float], train_loss: Dict[str, float],
+                  opt_pred: np.ndarray, valid_pred: np.ndarray,
```
LMZimmer: Is `opt_pred` simply `train_pred`?

autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -0,0 +1,533 @@
...
+        validation_loss, test_loss = self.calculate_auxiliary_losses(
+            valid_pred, test_pred, val_indices
+        )
+
+        if loss_ is not None:
+            return self.duration, loss_, self.seed, additional_run_info_
+
+        if isinstance(loss, dict):
+            loss_ = loss
+            loss = loss_[self.metric[0].name]
```
LMZimmer: Maybe it is less confusing to call it `optimization_loss` to not confuse it with the training loss. Also allowing multiple metrics and implicitly using the first one as loss function seems not that clean to me. Do you thing we could split that into `optimization_metric` and `additional_metrics` easily?

LMZimmer (commented):
Review Part 2 with some questions (doing the last part now)

autoPyTorch/utils/pipeline.py
ravinkohli: Its basically dataset_properties from the datamanager.

autoPyTorch/datasets/base_dataset.py
LMZimmer: Then let's just create an `else: raise NotImplementedError("")` explaining that

autoPyTorch/evaluation/tae.py
```diff
@@ -0,0 +1,426 @@
...
+def _encode_exit_status(exit_status: multiprocessing.connection.Connection) -> str:
+    try:
+        json.dumps(exit_status)
+        return exit_status
+    except (TypeError, OverflowError):
+        return str(exit_status)
+
+
+class ExecuteTaFuncWithQueue(AbstractTAFunc):
+
```
LMZimmer, line 90: Let's at least add a docstring for the purpose of this class, later we can add an explanation for the parameters

autoPyTorch/evaluation/abstract_evaluator.py
ravinkohli: I have added them in a new [commit](https://github.com/ravinkohli/Auto-PyTorch_refactor/blob/d6bc09b78ebf99e99278d91efdda4ca07c306a39/autoPyTorch/evaluation/abstract_evaluator.py#L166) but somehow it is not showing in this PR.

autoPyTorch/evaluation/abstract_evaluator.py
ravinkohli: actually I am also a bit unclear about what exactly it does, however, in the code, it saves Y_optimization to targets for the ensemble, see [here](https://github.com/ravinkohli/Auto-PyTorch_refactor/blob/d6bc09b78ebf99e99278d91efdda4ca07c306a39/autoPyTorch/evaluation/abstract_evaluator.py#L444)

autoPyTorch/evaluation/tae.py
```diff
@@ -0,0 +1,426 @@
...
+            logger_port=self.logger_port
+        )
+
+        try:
+            obj = pynisher.enforce_limits(**arguments)(self.ta)
+            obj(**obj_kwargs)
+        except Exception as e:
+            exception_traceback = traceback.format_exc()
+            error_message = repr(e)
+            additional_info = {
```
LMZimmer: The only critique for this module I have is that some variable names could be more specific, e.g. `info`, `additional_info`, `arguments`

autoPyTorch/evaluation/utils.py
```diff
@@ -0,0 +1,80 @@
...
+        except queue.Empty:
+            break
+
+    queue_.close()
+
+
+def extract_learning_curve(stack: List[RunValue], key: Optional[str] = None) -> List[List]:
+    learning_curve = []
+    for entry in stack:
+        if key:
```
LMZimmer: Seems like lazy coding for `if key is not None` to me :P

autoPyTorch/evaluation/train_evaluator.py
```diff
@@ -0,0 +1,340 @@
...
+    def _fit_and_predict(self, model: BaseEstimator, fold: int, train_indices: Union[np.ndarray, List],
+                         test_indices: Union[np.ndarray, List],
+                         add_model_to_self: bool
+                         ) -> Tuple[np.ndarray, np.ndarray, Optional[np.ndarray], Optional[np.ndarray]]:
+
+        self.indices[fold] = ((train_indices, test_indices))
+
+        # Do only output the files in the case of iterative holdout,
+        # In case of iterative partial cv, no file output is needed
+        # because ensembles cannot be built
```
franchuterivera: I think that in our case we can remove partial*

autoPyTorch/evaluation/train_evaluator.py
```diff
@@ -0,0 +1,340 @@
...
+
+        return Y_train_pred, Y_opt_pred, Y_valid_pred, Y_test_pred
+
+    def _predict(self, model: BaseEstimator,
+                 test_indices: Union[np.ndarray, List],
+                 train_indices: Union[np.ndarray, List]
+                 ) -> Tuple[np.ndarray, np.ndarray, Optional[np.ndarray], Optional[np.ndarray]]:
+
+        X_train = self.X_train
+        if isinstance(X_train, pd.DataFrame):
```
franchuterivera: I think we need to do something else here. we cannot just cast to numpy. This is something that the pipeline should handle. Casting to numpy will remove strings, that the pipeline might expect.

autoPyTorch/optimizer/smbo.py
```diff
@@ -0,0 +1,325 @@
...
+                 seed: int = 1,
+                 resampling_strategy: str = 'holdout',
+                 resampling_strategy_args: typing.Optional[typing.Dict[str, typing.Any]] = None,
+                 include: typing.Optional[typing.Dict[str, typing.Any]] = None,
+                 exclude: typing.Optional[typing.Dict[str, typing.Any]] = None,
+                 disable_file_output: bool = False,
+                 smac_scenario_args: typing.Optional[typing.Dict[str, typing.Any]] = None,
+                 get_smac_object_callback: typing.Optional[typing.Callable] = None,
+                 # TODO: Re-enable when ensemble merged
+                 # ensemble_callback: typing.Optional[EnsembleBuilderManager] = None,
```
franchuterivera, line 100: We can turn this on to see if everything is working. When you are done with the changes I can do a PR on your PR to enable this

autoPyTorch/pipeline/base_pipeline.py
franchuterivera: I think we probably want all of this in another file where to read the defaults in general

franchuterivera:
I recommend the following test to be added:

-  Test that we can fit a dummy classifier

- Test that we can fit a random forest with this and that the result is 1 (when evaluated with the same data used for training)

- Test that we can capture an error. Make a fit that crash so we see what happens when you crash a fit configuration

- We have to make sure the predictions saved to disk are for all the data.  check the size of the predictions saved by finish_up

autoPyTorch/evaluation/train_evaluator.py
```diff
@@ -0,0 +1,340 @@
...
+            status = StatusType.SUCCESS
+            self.logger.debug("In train evaluator fit_predict_and_loss, loss:{}".format(opt_loss))
+            self.finish_up(
+                loss=opt_loss,
+                train_loss=train_loss,
+                opt_pred=Y_optimization_preds,
+                valid_pred=Y_valid_preds,
+                test_pred=Y_test_preds,
+                additional_run_info=additional_run_info,
+                file_output=self.num_folds == 1,
```
LMZimmer: does this have to be adapted? if not this block is redundant with line 105 and could be placed outside the if else

LMZimmer (commented):
Last part. Also does auto-sklearn have tests for the code we use from them? If so let's copy them, otherwise Francisco's examples seem nice

autoPyTorch/pipeline/base_pipeline.py
LMZimmer: Yeah, probably autoPyTorch/configs would be most suitable, we can also store search space presets there. But thats not the most urgent point

autoPyTorch/evaluation/train_evaluator.py
ravinkohli: the error is in accessing the data via indices if it is a dataframe we need to call iloc on it rather than how it is done right now. Also, I tried casting to numpy and it didn't change the data frame, it just made an np array with dtype=object

autoPyTorch/optimizer/smbo.py
ravinkohli: Sure

autoPyTorch/utils/pipeline.py
ravinkohli: To get dataset requirements we need to instantiate the pipeline, so the idea of having this in the utils file is that we use it to just get dataset_requirements and the config space for the general pipeline and the model_class attribute is used to generate models when trying to fit and predict a pipeline. Also, I think we need to send the dataset properties to get the pipeline configuration space as it is altered according to dataset properties. But I think it should be possible to instantiate the model class just to get the requirements and config space and hence we can avoid the double-checking of task type as being done right now.

autoPyTorch/evaluation/train_evaluator.py
ravinkohli: the name of the variables are different and they also contain different info, in holdout, the y_*_pred are predictions on one fold, and in cross val its a concatenation of all the folds. I thought it might be confusing if I try to keep the finish_up outside of the if else

autoPyTorch/evaluation/abstract_evaluator.py
ravinkohli: I guess init_params when initialising the pipeline is different than the fit dictionary, I'll remove it from this call

autoPyTorch/evaluation/abstract_evaluator.py
ravinkohli: opt_pred is the predictions on the val split from the resampling strategy, and valid_pred is predictions on the validation set if provided by the user, same for test_pred

autoPyTorch/evaluation/abstract_evaluator.py
LMZimmer: I see, it is just a flag wether to save the labels/targets, let's write that somewhere :)

autoPyTorch/evaluation/train_evaluator.py
LMZimmer: Okay, resolving this then

autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -0,0 +1,533 @@
...
+            if not isinstance(self.configuration, Configuration):
+                self.pipeline_class = DummyClassificationPipeline
+            else:
+                if self.task_type in TABULAR_TASKS:
+                    self.pipeline_class = TabularClassificationPipeline
+                elif self.task_type in IMAGE_TASKS:
+                    self.pipeline_class = ImageClassificationPipeline
+                else:
+                    raise ValueError('task {} not available'.format(self.task_type))
+            self.predict_function = self._predict_regression
```
LMZimmer: I don't quite understand this naming `self._predict_regression` since it is used for regression and classification tasks and there is not `self._predict_classification`. Do we need another function or rename this one?

autoPyTorch/evaluation/train_evaluator.py
LMZimmer: Looks like `_predict_regression` just calls `pipeline.predict`. I would assume that that can handle whatever comes from the dataset class (`pd.DataFrame` or `np.ndarray`) and `X_train` here also comes from the dataset. So the issue is just in line 286/289. That we can handle with and if-else or sth like:

```
X_subsampler = lambda x: X_train[x] if isinstance(X_train, np.ndarray) else X_train.iloc[x]
train_pred = self.predict_function(X_subsampler(train_indices), ...)
```
?

autoPyTorch/evaluation/train_evaluator.py
```diff
@@ -0,0 +1,340 @@
...
+                 ) -> Tuple[np.ndarray, np.ndarray, Optional[np.ndarray], Optional[np.ndarray]]:
+
+        X_train = self.X_train
+        if isinstance(X_train, pd.DataFrame):
+            X_train = X_train.to_numpy()
+        train_pred = self.predict_function(X_train[train_indices], model,
+                                           self.y_train[train_indices])
+
+        opt_pred = self.predict_function(X_train[test_indices], model,
+                                         self.y_train[train_indices])
```
LMZimmer, line 285: this should be `self.y_train[test_indices]` no?

EDIT: Actually, `self._predict_regression` does not even use the Y input, so we can just drop it everywhere

autoPyTorch/evaluation/train_evaluator.py
```diff
@@ -0,0 +1,340 @@
...
+
+        if self.X_valid is not None:
+            valid_pred = self.predict_function(self.X_valid, model,
```
LMZimmer: It is not quite clear to me what the differences between `X_train[test_indices]` and `self.X_valid` and `X_test` (and therefore `opt_pred`, `valid_pred`, `test_pred) are. Could you double check that there is no redundancy here?

autoPyTorch/evaluation/abstract_evaluator.py
ravinkohli: yeah we can but we are doing it with 0 indexing as the get_metrics always returns a list of metrics even when we ask for a specific.

autoPyTorch/evaluation/train_evaluator.py
ravinkohli: yeah I think its just in the comments, but I have removed it

autoPyTorch/evaluation/train_evaluator.py
ravinkohli: Thanks for the suggestion, looks good to me, I'll make it this way then

autoPyTorch/evaluation/utils.py
ravinkohli: fixed :P

autoPyTorch/evaluation/abstract_evaluator.py
ravinkohli: yeah we need a predict_proba for classification, which we dont have so I created an issue for it and in the time being just using the preidct_regression as it just calls the model predict

autoPyTorch/evaluation/train_evaluator.py
ravinkohli: when we have this predict_proba function, then we need the Y_train to ensure prediction_array_sizes, therefore I left it for now

autoPyTorch/evaluation/train_evaluator.py
ravinkohli: self.X_valid is coming from the datamanager, so in case the user instantiates the dataset by giving a validation set, and a test set then we'll have self.X_valid and self.X_test not None. X_train[test_indices] comes from creating splits in the dataset class, so basically from our resampling strategy.

ravinkohli:
> I recommend the following test to be added:
> 
>     * Test that we can fit a dummy classifier
> 
>     * Test that we can fit a random forest with this and that the result is 1 (when evaluated with the same data used for training)
> 
>     * Test that we can capture an error. Make a fit that crash so we see what happens when you crash a fit configuration
> 
>     * We have to make sure the predictions saved to disk are for all the data.  check the size of the predictions saved by finish_up

Your second point means that we can fit a random forest using the ExecuteTaeFuncWithQueue, without having smac or the evaluator, right?

autoPyTorch/evaluation/abstract_evaluator.py
ravinkohli: I have changed the code such as there is a flag called all_supported_metrics, which if true, we'll have additional_metrics filled with all the metrics supported for the task and metric will still be entered through the automlsmbo which will act as the optimization metric.

franchuterivera:
>Your second point means that we can fit a random forest using the ExecuteTaeFuncWithQueue, without having smac or the evaluator, right?

Yes, so that we make sure that your code in isolation works no matter what.

franchuterivera (approved):
Because all comments have been addressed, I will merge this PR. If I see an issue that was not captured by unit testing while implementing the automl/estimator classes I will create an issue

PR #60, opened 2020-12-23
BaseTask and TabularClassificationTask

### LINKED ISSUE #39: Enhance logging via a logging server
We found out in autosklearn that when doing distributed computing, the generation of task in a worker does not retain the root logger.

A solution to this problem is to create a logging server. This is to remind us to port over the logging server once it is integrated to autosklearn.

### LINKED ISSUE #57: Predict Probability doesn't work on numerical only dataset -- because of multiclass classification
This problem can be easily reproduced by removing this [line](https://github.com/LMZimmer/Auto-PyTorch_refactor/blob/df358bcfc04a88615ffa12f2c5b8bcccfe5e2590/test/test_pipeline/test_tabular_classification.py#L61)

The problem is that https://github.com/LMZimmer/Auto-PyTorch_refactor/blob/3683897e3587971d50a2fcf82897c0a65b45e927/autoPyTorch/pipeline/tabular_classification.py#L114 assumes that the output shape is unidimensional or not a tuple.

The goal is to enable predict probability in this testcase.

ravinkohli:
This PR contains the following-
1. BaseTask with 
    a. fit- Runs smac on the given dataset and runs ensemble builder to create ensembles of the best pipelines found.
    b. predict- Gives ensemble predictions on the given dataset.
    c. refit- Refits previously found pipelines and ensembles on a different dataset and new budget can also be defined. 
    d. score- Score on test set
2. TabularClassificationTask
3. Corrections made to hyperparameter search space in various components to optimise performance.
4. Fixes bug where networks were not properly pickled and hence affected performance.
5. Adds logging server. resolves #39 
6. Resolves #57


autoPyTorch/api/default_pipeline_options.json
```diff
@@ -0,0 +1,11 @@
+{
```
LMZimmer: Could we maybe move these to a new directory autoPyTorch/configs ?

autoPyTorch/api/default_pipeline_options.json
```diff
@@ -0,0 +1,11 @@
...
+            "device": "cpu",
+            "budget_type": "epochs",
+            "epochs": 100,
```
LMZimmer: I think 50 is plenty

autoPyTorch/api/base_task.py
```diff
@@ -1,139 +1,740 @@
...
-                generate the splits based on different restrictions.
+            name (str): name of the log file,
+            usually the dataset name
+
+        Returns:
+            PicklableClientLogger
         """
-        # TODO: Check dataset type against task type
-        dataset_properties = dataset.get_dataset_properties()
+        logger_name = 'AutoML:%s' % name
```
LMZimmer: Better `AutoPyTorch:%s` ?

autoPyTorch/api/base_task.py
```diff
@@ -1,139 +1,740 @@
...
+        self._logger_port = logging.handlers.DEFAULT_TCP_LOGGING_PORT
+
+        # Store the resampling strategy from the dataset, to load models as needed
+        self.resampling_strategy = None  # type: Optional[Union[CrossValTypes, HoldoutValTypes]]
 
     @abstractmethod
-    def build_pipeline(self):
+    def _get_required_dataset_properties(self, dataset: BaseDataset) -> Dict[str, Any]:
+        """
+        using the dataset entered, this function will
```
LMZimmer: This should rather be "given a pipeline type, this function returns the dataset properties required by the dataset object"

autoPyTorch/api/base_task.py
```diff
@@ -1,139 +1,740 @@
...
+        self._disable_file_output = disable_file_output
+        # Save start time to backend
+        self._backend.save_start_time(str(self.seed))
+
+        self._backend.save_datamanager(dataset)
+
+        self._metric = get_metrics(
+            names=[optimize_metric], dataset_properties=dataset_properties)[0]
+
+        self.search_space = get_configuration_space(info=dataset_properties,
```
LMZimmer: I think this should be `self.search_space = self.get_search_space(...)` and `self.get_search_space` should be something like:

```
def get_search_space(self, dataset=None):
    if self.search_space is not None:
        return self.search_space
    elif dataset is not None:
        return get_configuration_space(info=dataset.get_dataset.properties, include=self.include_components, ...)
    raise Exception(...)
```

Such that `fit` uses `self.search_space` in case the user has set it via `.set_search_space()`

autoPyTorch/api/base_task.py
```diff
@@ -1,139 +1,740 @@
...
+                                                    include=self.include_components,
+                                                    exclude=self.exclude_components)
+        budget_config: Dict[str, Union[float, str]] = {}
+        if budget_type is not None and budget is not None:
+            budget_config['budget_type'] = budget_type
+            budget_config[budget_type] = budget
+        elif budget_type is not None or budget is not None:
+            raise ValueError(
+                "budget was mentioned but not the budget_type to be used with it"
```
LMZimmer: Maybe better: "budget_type was not specified in budget_config"

autoPyTorch/api/base_task.py
```diff
@@ -1,139 +1,740 @@
...
+            )
+
+        if self.task_type is None:
+            raise ValueError("Cannot interpret task type from the dataset")
+
+        self._create_dask_client()
+        proc_ensemble = None
+        if self.ensemble_size <= 0:
```
LMZimmer, line 587: Couldn't this be <=1  ?

autoPyTorch/api/base_task.py
```diff
@@ -1,139 +1,740 @@
...
+                "No valid ensemble was created. Please check the log"
+                "file for errors. Default to the best individual estimator:{}".format(
+                    ensemble.identifiers_
+                )
+            )
+
+        return ensemble
 
-    @typing.no_type_check
     def fit(
```
LMZimmer: Can we call this `search`? The fit-refit thing confused everyone

autoPyTorch/api/base_task.py
```diff
@@ -1,139 +1,740 @@
...
 
-        X = {}
-        X.update(dataset_properties)
-        X.update(self.pipeline_config)
+            # try to fit the model. If it fails, shuffle the data. This
+            # could alleviate the problem in algorithms that depend on
+            # the ordering of the data.
+            fit_and_suppress_warnings(self._logger, model, X, y=None)
 
-        self.pipeline.set_hyperparameters(model_config)
```
LMZimmer, line 129: Could we add a new method that takes a dataset and a model config (a sample from a ConfigSpace object) and trains the pipeline? Basically what l. 129/130 does. That is a pretty useful feature if you just want to evaluate one config as a user. Maybe call that one `.fit` or `.train`

autoPyTorch/api/base_task.py
```diff
@@ -143,40 +744,100 @@ def predict(
...
-        # normally and raise warning
-        return self.pipeline.predict(X_test)
+
+        # Parallelize predictions across models with n_jobs processes.
+        # Each process computes predictions in chunks of batch_size rows.
+        if self._logger is None:
+            self._logger = self._get_logger("Predict-Logger")
+
+        if self.ensemble_ is None and not self._load_models(self.resampling_strategy):
+            raise ValueError("Failed to fit an ensemble. Cannot predict")
```
LMZimmer: Maybe "No ensemble found. Either fit has not yet been called or no ensemble was fitted"

autoPyTorch/api/base_task.py
```diff
@@ -143,40 +744,100 @@ def predict(
...
+            raise ValueError('Something went wrong generating the predictions. '
+                             'The ensemble should consist of the following '
+                             'models: %s, the following models were loaded: '
+                             '%s' % (str(list(self.ensemble_.indices_)),
+                                     str(list(self.models_))))
+
+        predictions = self.ensemble_.predict(all_predictions)
+
+        if self.task_type not in REGRESSION_TASKS:
+            # Make sure average prediction probabilities
```
LMZimmer: Better delete the "average"

autoPyTorch/api/base_task.py
```diff
@@ -143,40 +744,100 @@ def predict(
...
+            # are within a valid range
+            # Individual models are checked in _model_predict
+            predictions = np.clip(predictions, 0.0, 1.0)
```
LMZimmer: This could be dangerous if there are errors in the voter and make debugging hard. Maybe better throw an exception or does this actually occur legally in some cases?

autoPyTorch/api/base_task.py
```diff
@@ -143,40 +744,100 @@ def predict(
...
         Returns:
-            Value of the evaluation metric calculated on the test set.
+            Dict[str, float]: Value of the evaluation metric calculated on the test set.
         """
-        return self.pipeline.score(X_test, y_test, sample_weights)
+        if isinstance(y_test, pd.Series):
+            y_test = y_test.to_numpy(dtype=np.float)
+
+        if self._metric is None:
+            raise ValueError("AutoPytorch failed to infer a metric from the dataset "
```
LMZimmer: More precise "No metric found. Either fit/search has not been called yet or AutoPyTorch failed to infer a metric from the dataset"

autoPyTorch/api/base_task.py
```diff
@@ -143,40 +744,100 @@ def predict(
...
+                             "Please check the log file for related errors. ")
+        if self.task_type is None:
+            raise ValueError("AutoPytorch failed to infer a task type from the dataset "
```
LMZimmer, line 924: similar

autoPyTorch/api/base_task.py
```diff
@@ -143,40 +744,100 @@ def predict(
...
+                             "Please check the log file for related errors. ")
+        return calculate_score(target=y_test, prediction=y_pred,
+                               task_type=STRING_TO_TASK_TYPES[self.task_type],
+                               metrics=[self._metric])
 
     @typing.no_type_check
     def get_incumbent_results(
-        self
+            self
```
LMZimmer, line 932: This should return the score of the best ensemble/model

autoPyTorch/api/base_task.py
```diff
@@ -143,40 +744,100 @@ def predict(
...
     ):
         pass
 
     @typing.no_type_check
     def get_incumbent_config(
-        self
+            self
```
LMZimmer, line 938: This could return the configs of the ensemble models + their weights

autoPyTorch/api/base_task.py
```diff
@@ -1,139 +1,740 @@
...
                 unknown_keys.append(option)
 
         if len(unknown_keys) > 0:
-            raise ValueError("Invalid configuration arguments given %s" % unknown_keys)
+            raise ValueError("Invalid configuration arguments given {},"
+                             " expected arguments to be in {}".
+                             format(unknown_keys, self.default_pipeline_options.keys()))
 
-        self.pipeline_config.update(pipeline_config_kwargs)
+        self.default_pipeline_options.update(pipeline_config_kwargs)
```
LMZimmer: Probably should rename `self.default_pipelie_options` to `self.pipeline_options` since this is not the default anymore?

autoPyTorch/api/base_task.py
```diff
@@ -1,139 +1,740 @@
...
+            ensemble_size: int = 50,
+            ensemble_nbest: int = 50,
+            max_models_on_disc: int = 50,
+            temporary_directory: Optional[str] = None,
+            output_directory: Optional[str] = None,
+            delete_tmp_folder_after_terminate: bool = True,
+            delete_output_folder_after_terminate: bool = True,
+            include_components: Optional[Dict] = None,
+            exclude_components: Optional[Dict] = None,
+            backend: Optional[Backend] = None,
```
LMZimmer, line 135: Do you think there is a way to reduce the number of arguments here and in `fit`? E.g. could we pack all pipeline options to the default pipeline config (isn't seed for instance a pipeline option?). My worry is that we have a lot flying around with the arguments from `fit`, `init` and `self.default_pipeline_options` and that things dont get updates/used properly

autoPyTorch/datasets/tabular_dataset.py
```diff
@@ -58,6 +82,15 @@ def __init__(self, X: Any, Y: Any,
                  train_transforms: Optional[torchvision.transforms.Compose] = None,
                  val_transforms: Optional[torchvision.transforms.Compose] = None,
                  ):
+
+        # TODO: Remove. Here for debug. If still here, it means
+        # that I forgot to clean this, but we need to discuss about
+        # standard scaling due to exploding gradient
```
LMZimmer: what is the issue here?

autoPyTorch/evaluation/abstract_evaluator.py
```diff
@@ -249,28 +250,29 @@ def __init__(self, backend: Backend,
...
+        self.logger = get_named_client_logger(
+            name=logger_name,
+            port=logger_port,
+        )
         self.Y_optimization: Optional[np.ndarray] = None
         self.Y_actual_train: Optional[np.ndarray] = None
         self.pipelines: Optional[List[BaseEstimator]] = None
         self.pipeline: Optional[BaseEstimator] = None
-        self.logger.debug("Fit dictionary in Abtract evaluator: {}".format(self.fit_dictionary))
+        # self.logger.debug("Fit dictionary in Abtract evaluator: {}".format(self.fit_dictionary))
```
LMZimmer: Is this too much output? I think for debug we want as much as possible

setup.py
```diff
@@ -57,4 +57,5 @@
         "docs": ["sphinx", "sphinx-gallery", "sphinx_bootstrap_theme", "numpydoc"],
     },
     test_suite="pytest",
+    data_files=[('api', ['autoPyTorch/api/default_pipeline_options.json'])]
```
LMZimmer: Reminder to change this if we create `autoPyTorch/configs`

autoPyTorch/utils/hash.py
```diff
@@ -0,0 +1,29 @@
+import hashlib
+
+import numpy as np
+
+import scipy.sparse
+
+
+def hash_array_or_matrix(X: np.ndarray) -> str:
```
LMZimmer: This could probably also go to utils/common.py

LMZimmer (commented):
Thank you for the work, I left some ideas mainly in the base_task

autoPyTorch/api/base_task.py
franchuterivera: @ravinkohli  originally had search here, but I recommended him to call it fit, because TPOT/autosklearn/autogluon/Gamma/... uses this interface, so look like almost everyone uses this interface.  But if you got a lot of complains for this nothing more to say :)

autoPyTorch/api/base_task.py
franchuterivera: This is  handel a bit differently (see [this](https://github.com/LMZimmer/Auto-PyTorch_refactor/blob/4e2a1b507e930acec4afe03d2f85a39693f4ebea/autoPyTorch/api/base_task.py#L409)). So in this case we don't want to spend precious iterations doing ensemble selection when we already know which is the best model from smac. It is best to lest the above function select the best individual model in case ensemble size is 1 (My argument is also follows from the fact that ensemble selection calculates the score of the predictions which is the most time-consuming operation in the process, something we should consider addressing in the future, because of the many checks done in sklearn).

LMZimmer: perfect :)

autoPyTorch/api/base_task.py
LMZimmer: Yeah, I am actually wondering wether they though about it or just copied it from previous frameworks.. Anyways, `fit` is overloaded, in our case especially since we also have `refit` where fit refers to fitting a specific pipeline and not running the AutoML optimization. I'd say we either call this one `search` or rename `refit` to e.g. `retrain`. I am fine with either option

autoPyTorch/api/default_pipeline_options.json
ravinkohli: sure

autoPyTorch/api/base_task.py
ravinkohli: okay, I'll change it to search, so that fit or refit means fitting some pipeline on the given dataset.

autoPyTorch/api/base_task.py
ravinkohli: It was happening for some of the runs, but that was in individual pipelines where an error is raised in _pipeline_predict. But yeah I think raising an error might be better, so I'll do that here as well.

autoPyTorch/api/base_task.py
ravinkohli: Actually this task_type is not being inferred from the dataset, It's being decided per task, as we will have a separate class for each task_type and therefore it should never be None, I think this check is only there for mypy

autoPyTorch/evaluation/abstract_evaluator.py
ravinkohli: yeah It was printing the whole X_train and y_train multiple times over, but I guess fit dictionary is important for debug, so, I'll put everything except the dataset keys to be printed

autoPyTorch/api/base_task.py
ravinkohli: I agree with seed being a pipeline option but it's also used in smac and ensemble. And the other options all specify attributes which are used for the whole experiment and not in the pipeline config. All I can think of is maybe removing budget and budget type, but for that I feel, its a good option to allow just for search or refit a different budget.

autoPyTorch/datasets/tabular_dataset.py
ravinkohli: Actually I didnt add this, @franchuterivera did. But it seems I should delete it. When we finished the API the performance was not so great so maybe he added it to test if this could be a reason.

autoPyTorch/api/base_task.py
ravinkohli: I agree. Changed it

autoPyTorch/api/base_task.py
LMZimmer: Okay then let's keep it as is

LMZimmer (approved):

autoPyTorch/datasets/tabular_dataset.py
franchuterivera: As the comment says, I forgot to remove this :), can you please remove this?

requirements.txt
```diff
@@ -1,18 +1,25 @@
 pandas
 torch
 torchvision
-scikit-learn>=0.22.0,<0.23
+scikit-learn
```
franchuterivera: I think we should limit out version of scikit learn -- because there are normally very big changes in the latest versions

requirements.txt
```diff
@@ -1,18 +1,25 @@
...
+imgaug
 
 
-ConfigSpace>=0.4.14,<0.5
-pynisher>=0.6.3
+ConfigSpace
+pynisher
 pyrfr>=0.7,<0.9
-smac>=0.13.1,<0.14
+smac
```
franchuterivera: Same with smac/pynisher

requirements.txt
```diff
@@ -1,18 +1,25 @@
...
 dask
 distributed>=2.2.0
+
+pytest
```
franchuterivera: This is ot needed here right? And I do not thing sklearn is a package?

franchuterivera (changes requested):
Some minor comments from my side.... after that we should merge!

PR #61, opened 2020-12-29
Incorporate traditional ML algos for tabular

LMZimmer:
- Introduces a new traditional ML pipeline for tabular data
- Adds RF, LGB, KNN, ExtraTrees, ... to the traditional models

To be merged after the automl PR


autoPyTorch/api/base_task.py
```diff
@@ -492,7 +496,7 @@ def _do_dummy_prediction(self):
             logger=self._logger,
             cost_for_crash=get_cost_of_crash(self._metric),
             abort_on_first_run_crash=False,
-            initial_num_run=self._num_run,
+            initial_num_run=self._num_run if self._num_run is not None else 1,
```
franchuterivera: Ho can self._num_run be not 1? Do you think it is better to make num_run an argument to dummy prediction?

ravinkohli: I made it this way as in traditional classification the num_run depends on the number of available classifiers, which could be dynamic, therefore, this attribute of the class gets updated after it is run. and this if statement is to comply with mypy

autoPyTorch/api/base_task.py
```diff
@@ -542,8 +549,9 @@ def _do_traditional_prediction(self, time_for_traditional):
             memory_limit = int(math.ceil(memory_limit))
         available_classifiers = get_available_classifiers()
         dask_futures = list()
-        time_for_traditional_classifier_sec = int(time_for_traditional/len(available_classifiers))
-        for num_run, classifier in enumerate(available_classifiers, start=self._num_run+1):
+        time_for_traditional_classifier_sec = int(time_for_traditional / len(available_classifiers))
+        for num_run, classifier in enumerate(available_classifiers,
+                                             start=self._num_run + 1 if self._num_run is not None else 1):
```
franchuterivera: same here I think... probably it is more intuitive to just have num_run in argument. Return it to the main code and increase it as needed.

ravinkohli: okay I'll do this then

autoPyTorch/api/base_task.py
```diff
@@ -542,8 +549,9 @@ def _do_traditional_prediction(self, time_for_traditional):
...
             start_time = time.time()
```
franchuterivera, line 553: Can you also call:
```
 self._stopwatch.start_task(ensemble_task_name)
```
Like call the stop watch to track how much this task take

autoPyTorch/api/base_task.py
```diff
@@ -1093,7 +1088,7 @@ def score(
     def __getstate__(self) -> Dict[str, Any]:
         # Cannot serialize a client!
         self._dask_client = None
-        self.logging_server = None
+        self.logging_server = None  # type: ignore [assignment]
```
franchuterivera, line 1092: Isn;t it better to define  self.logging_server  as an Optional?

examples/example_tabular_classification.py
```diff
@@ -43,7 +44,9 @@ def get_data_to_train() -> typing.Tuple[typing.Any, typing.Any, typing.Any, typi
         X=X_train, Y=y_train,
         X_test=X_test, Y_test=y_test)
 
-    api = TabularClassificationTask(delete_tmp_folder_after_terminate=False)
+    api = TabularClassificationTask(delete_tmp_folder_after_terminate=False,
+                                    temporary_directory='./tmp/example_tabular_classification_tmp',
```
franchuterivera: Can you change back to not using a provided directory? For the examples it make sense to try the most common behaviour, which is not providing... (During unit testing we test providing a path, so this is like the only place we test not having this temporay_directory=<> flag)

autoPyTorch/api/base_task.py
ravinkohli: in one of the comments you have mentioned that, "BaseContext" has no attribute "Process" motivates to ignore the attr check, therefore I thought to ignore the assignment

examples/example_tabular_classification.py
ravinkohli: yeah sure

test/test_pipeline/components/test_setup_traditional_classification.py
```diff
@@ -0,0 +1,128 @@
...
+        assert isinstance(results['val_score'], float)
+
+        # Test if classifier can predict on val set and
+        # if the result is same as the one in results
+        y_pred = classifier.predict(X, predict_proba=True)
+        assert np.allclose(y_pred, results['val_preds'], atol=1e-04)
+
+        # Test if classifier can score and
+        # the result is same as in results
+        score = classifier.score(X, y)
```
franchuterivera, line 122: can you add a check that score should be greater than 0.8 ? because ideally this is super easy data to fit

test/test_pipeline/components/test_setup_traditional_classification.py
```diff
@@ -0,0 +1,128 @@
...
+        assert 'labels' in results.keys()
+        assert len(results['labels']) == y.shape[0]
+        assert 'train_score' in results.keys()
+        assert isinstance(results['train_score'], float)
+        assert 'val_score' in results.keys()
+        assert isinstance(results['val_score'], float)
+
+        # Test if classifier can predict on val set and
+        # if the result is same as the one in results
+        y_pred = classifier.predict(X, predict_proba=True)
```
franchuterivera, line 117: can you add a check that y_pred has the same number of samples as X, y

autoPyTorch/api/base_task.py
franchuterivera: Oh, ok I see, then never mind. This BaseContext is something we cannot deal with.

test/test_pipeline/components/test_setup_traditional_classification.py
ravinkohli: sure

test/test_pipeline/components/test_setup_traditional_classification.py
ravinkohli: sure

autoPyTorch/api/base_task.py
```diff
@@ -583,16 +724,47 @@ def search(
...
+
+        traditional_task_name = 'runTraditional'
+        self._stopwatch.start_task(traditional_task_name)
+        elapsed_time = self._stopwatch.wall_elapsed(self.dataset_name)
+        time_for_traditional = int(traditional_per_total_budget * max(0, (self._time_for_task - elapsed_time)))
+        if time_for_traditional <= 0:
+            if traditional_per_total_budget > 0:
+                raise ValueError("Not enough time allocated to run traditional algorithms")
+        elif traditional_per_total_budget != 0:
+            num_run = self._do_traditional_prediction(num_run=num_run, time_for_traditional=time_for_traditional)
```
franchuterivera: Shouldn't this be:
```
 num_run = self._do_traditional_prediction(num_run=num_run+1, time_for_traditional=time_for_traditional)
```

The num_run+1  to account for dummy predictions?

test/test_pipeline/components/test_setup_traditional_classification.py
```diff
@@ -117,12 +117,13 @@ def test_classifier_fit_predict(self, classifier, dataset):
...
         # Test if classifier can score and
         # the result is same as in results
         score = classifier.score(X, y)
         assert score == results['val_score']
-
+        # Test if score is greater than 0.8
+        assert score >= 0.8
         # delete folder created by catboost
         if isinstance(classifier, CatboostModel):
             shutil.rmtree('./catboost_info')
```
franchuterivera: BTW, this is a super important change. We do not have to do it now, but eventually. I am struggling with the automl autopytorch because of this. Files get created in the current directory. They should be created in backend. This is just an FYI that we will have to change this in the future.

autoPyTorch/api/base_task.py
ravinkohli: yeah sorry.

test/test_pipeline/components/test_setup_traditional_classification.py
franchuterivera: actually if this is not a lot of rouble, this seems like super easy to fix.

One has to just add:

https://catboost.ai/docs/search/?query=train_dir


Meaning, can you please here https://github.com/LMZimmer/Auto-PyTorch_refactor/blob/45901a53327a1e982bf8e1ae09d2f182d9348e01/autoPyTorch/pipeline/components/setup/traditional_ml/classifier_models/classifiers.py#L112 add this line:

https://github.com/franchuterivera/Auto-PyTorch/blob/31381aee6557ebe184f4db42e62300e81edb2983/autoPyTorch/components/baselines/baselines.py#L162

I THINK that should work? so we can remove line 129 and no file gets created at top level. 


This is super important because singularity in our cluster do no allow files to be created in a read only file system.

